PR #54, opened 2019-04-24
Propagate user-provided where clauses

shepmaster:


mikeando:
Does this work with enums like:

~~~
#[derive(Debug, Snafu)]
enum Error<T:std::fmt::Display> {
     #[snafu(display("The value {} was too large", value))]
     TooLarge { value: T, limit: u32 },

     #[snafu(display("The value {} was too small", value))]
     TooSmall { value: T, limit: u32 },
}
~~~

I'm guessing its OK, but don't have the tools to check it available to me right now...

shepmaster:
> I'm guessing its OK

I was *also* guessing it was ok, but that's no way to live. Actually **testing it** shows that it doesn't 😊 ! I'm going to open up a new issue to fix that, and then do a combined release.

Issue #58, opened 2019-04-24
Consider automatically using the source error when `display` isn't provided.


shepmaster:
```rust
#[derive(Snafu)]
enum Error {
    Something { source: io::Error }
}
```

Is now effectively `#[snafu(display("Something"))]`, but we could do `#[snafu(display("Something: {}", source))]`.

shepmaster:
Closed via #63

PR #63, opened 2019-04-27
Default display attribute includes the source error description

dimfeld:
For #58. It looked like you weren't quite sure if you actually want to do this but I've wanted to check out procedural macros and this seemed like an easy intro. :) The test still needs a lot of improvement, which I can probably get to next week.

### DIFF
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -669,9 +669,14 @@ impl<'a> DisplayImpl<'a> {
                     ..
                 } = *variant;
 
-                let format = display_format
-                    .as_ref()
-                    .map_or_else(|| quote! { stringify!(#variant_name) }, |v| quote! { #v });
+                let format = match (display_format, source_field) {
+                    (&Some(ref v), _) => quote! { #v },
+                    (&None, &Some(ref f)) => {
+                        let field_name = &f.name;
+                        quote! { concat!(stringify!(#variant_name), ": {}"), #field_name }
+                    }
+                    (&None, &None) => quote! { stringify!(#variant_name)},
+                };
 
                 let field_names = user_fields
                     .iter()
--- a/src/guide/the_macro.rs
+++ b/src/guide/the_macro.rs
@@ -124,7 +124,8 @@
 //! ```
 //!
 //! If no display format is specified, the variant's name will be used
-//! by default.
+//! by default. If the field is an underlying error, that error's
+//! `Display` implementation will also be included.
 //!
 //! #### `ErrorCompat`
 //!
--- a/tests/default_error_display.rs
+++ b/tests/default_error_display.rs
@@ -1,0 +1,20 @@
+extern crate snafu;
+
+use snafu::{ResultExt, Snafu};
+
+#[derive(Debug, Snafu)]
+enum InnerError {
+    #[snafu(display("inner error"))]
+    AnError,
+}
+
+#[derive(Debug, Snafu)]
+enum Error {
+    NoDisplay { source: InnerError },
+}
+
+#[test]
+fn default_error_display() {
+    let err: Error = AnError.fail::<()>().eager_context(NoDisplay).unwrap_err();
+    assert_eq!(format!("{}", err), "NoDisplay: inner error",);
+}


snafu-derive/src/lib.rs
```diff
@@ -669,9 +669,19 @@ impl<'a> DisplayImpl<'a> {
...
+                            let field_name = &source_field.name;
+                            // Format before using quote! to avoid adding extra whitespace between the
+                            // variant name and the colon.
+                            let format_string = format!("{}: {{}}", variant_name);
+                            quote! { #format_string, #field_name }
+                        }
+                        None => quote! { stringify!(#variant_name) },
+                    },
+                    |v| quote! { #v },
+                );
```
shepmaster: This section might flow better as (pseudocode)

```rust
match (display_format, source_field) {
    (Some(v), _) => ...,
    (None, Some(f)) => ...,
    (None, None) => ...,
}
```

Since we support Rust 1.18, you will probably need to use the explicit dereferencing and `ref` keyword as appropriate.

snafu-derive/src/lib.rs
```diff
@@ -669,9 +669,19 @@ impl<'a> DisplayImpl<'a> {
...
-                let format = display_format
-                    .as_ref()
-                    .map_or_else(|| quote! { stringify!(#variant_name) }, |v| quote! { #v });
+                let format = display_format.as_ref().map_or_else(
+                    || match source_field {
+                        Some(source_field) => {
+                            let field_name = &source_field.name;
+                            // Format before using quote! to avoid adding extra whitespace between the
+                            // variant name and the colon.
+                            let format_string = format!("{}: {{}}", variant_name);
```
shepmaster: I think you can avoid the allocation and escaping the `{}` by using `concat!`:

```rust
concat!(stringify!(#variant_name), ": {}")
```

tests/error_message.rs
```diff
@@ -0,0 +1,17 @@
+extern crate snafu;
+
+use std::error::Error as _;
```
shepmaster: I'm betting the `as _` won't work in the older Rust versions, but I don't think you need it either way.

tests/error_message.rs
```diff
@@ -0,0 +1,17 @@
...
+
+
+#[derive(Debug, Snafu)]
+enum Error {
+    BadId{ source: std::io::Error },
+}
+
+#[test]
+fn test_error_message() {
+    let err : Error = std::fs::read("/non/existent/file--test").eager_context(BadId{}).unwrap_err();
```
shepmaster: For consistency, I'd call the context selector as `eager_context(BadId)`.

shepmaster: I don't think `fs::read` exists in Rust 1.18, but you can use `File::open` for the same effect.

tests/error_message.rs
```diff
@@ -0,0 +1,17 @@
...
+    let err_description = format!("{}", err);
+    assert_eq!(err_description, String::from("BadId: No such file or directory (os error 2)"));
```
shepmaster: You shouldn't need to allocate the `String` to test against; the string literal is fine.

shepmaster:
I worry a little bit that the error message from `io::Error` *might* be platform-specific. It also relies on that file not being there. Combined with the comment about `fs::read` in Rust 1.18, it might not be the best choice for this test.

Another option, instead of relying on `io:Error`, is to use SNAFU itself to make the underlying error:

```rust
#[derive(Debug, Snafu)]
enum Inner {
    #[snafu(display("This is the inner error"))]
    Something,
}

#[derive(Debug, Snafu)]
enum Error {
    NoDisplay { source: Inner },
}
```

```rust
Something.fail().eager_context(NoDisplay).unwrap_err();
```

This gives the test complete control over what it wants to check.

tests/error_message.rs
dimfeld: Oh great, I didn't realize you could omit the brackets in this case.

dimfeld:
Yeah I agree completely about the use of `fs::read`, and it makes a lot of sense to use another Snafu as the source. 

I think this latest commit should address everything you mentioned. Thanks for the detailed feedback, and thanks for making this crate! It's exactly what I wanted for managing error flow.

dimfeld:
Squashed. Thanks!

PR #72, opened 2019-05-08
Implement extension traits for futures 0.1

### LINKED ISSUE #41: Add extension traits for futures
These can be behind feature flags.

Should support both the nightly-only Futures and the crate.

ErichDonGubler:
Currently, this PR is a WIP because this:
 * currently only targets the `futures` crate. However, [Rust nightly's `Future`](https://doc.rust-lang.org/nightly/core/future/trait.Future.html) should be REALLY similar (i.e., only applying `where Self::Output: Result<T, E>`).
* doesn't yet attempt to use Cargo features to switch things on and off.
* doesn't add any additional tests, which may or may not actually be a deal breaker for merging. :)

These things haven't been done yet because after testing a bit it seems that the non-`eager` methods might not be as useful, and I wanted to ask @shepmaster's opinion on the API design before continuing. Docs might need to be tweaked since I didn't put much brainpower into it beyond adapting for the different set of types that are being used for `FutureExt`.


Cargo.toml
```diff
@@ -42,3 +42,4 @@ exclude = [
 [dependencies]
 snafu-derive = { path = "snafu-derive", version = "0.2.3" }
 backtrace = { version = "0.3.0", optional = true }
+futures = "0.1.26"
```
shepmaster: For the widest compatibility, we want to make this as permissive as possible, which is *likely* to be all the way back to just `0.1`.

src/lib.rs
```diff
@@ -430,6 +432,311 @@ impl<T> OptionExt<T> for Option<T> {
     }
 }
 
+/// Additions to [`Future`](futures::Future).
```
shepmaster: To avoid having to put a bunch of stuff in `cfg` attributes, this code will likely want to be put into a module (something like `futures_crate`?) and then we can have a minimal number of `cfg`s: one for the crate, one for the module, and one for the re-export of the trait.

src/lib.rs
```diff
@@ -430,6 +432,311 @@ impl<T> OptionExt<T> for Option<T> {
...
+    /// use futures::Future;
+    /// # use futures::IntoFuture;
+    /// use snafu::{Context, FutureExt, Snafu};
+    ///
+    /// #[derive(Debug, Snafu)]
+    /// enum Error {
+    ///     Authenticating { user_name: String, user_id: i32, source: ApiError },
+    /// }
+    ///
+    /// fn example() -> impl Future<Item=i32, Error=Context<ApiError, Authenticating<&'static str, i32>>> {
```
shepmaster:
```suggestion
    /// fn example() -> impl Future<Item = i32, Error = Context<ApiError, Authenticating<&'static str, i32>>> {
```

shepmaster: Ditto this elsewhere. Someday I'll figure out how to run rustfmt on code examples, but for now it's all by hand...

src/lib.rs
```diff
@@ -430,6 +432,311 @@ impl<T> OptionExt<T> for Option<T> {
...
+    type Item = A::Item;
+    type Error = Context<A::Error, C>;
+
+    fn poll(&mut self) -> Result<futures::Async<Self::Item>, Self::Error> {
+        self.future.poll().map_err(|error| Context {
+            error,
+            context: self
+                .context
+                .take()
+                .expect("cannot poll ContextFuture twice"),
```
shepmaster: I know it's part of futures, but hate having to do this. 😿

src/lib.rs
```diff
@@ -430,6 +432,311 @@ impl<T> OptionExt<T> for Option<T> {
...
+    }
+}
+
+/// Future for the `with_context` combinator, changing the error type of a future.
+///
+/// See the `FutureExt::with_context` method for more details.
+pub struct WithContextFuture<A, C, F: FnOnce() -> C> {
+    context: Option<F>,
+    future: A,
+    _c: std::marker::PhantomData<C>,
```
shepmaster: I'm surprised by this; I'll have to compile locally to understand why it's needed.

src/lib.rs
```diff
@@ -430,6 +432,311 @@ impl<T> OptionExt<T> for Option<T> {
...
+    type Error = Context<A::Error, C>;
+
+    fn poll(&mut self) -> Result<futures::Async<Self::Item>, Self::Error> {
+        self.future.poll().map_err(|error| Context {
+            error,
+            context: (self
+                .context
+                .take()
+                .expect("cannot poll WithContextFuture twice"))(),
+        })
```
shepmaster: I'd probably flatten these out a little:

```rust
        self.future.poll().map_err(|error| {
            let context = self
                .context
                .take()
                .expect("cannot poll WithContextFuture twice");
            let context = context();

            Context { error, context }
        })
```

Cargo.toml
BurntSushi: Could we also make `futures` an optional dep?

ErichDonGubler: @BurntSushi: That's the plan -- I just haven't formally added any associated features yet, so that simply hasn't been done yet. :)

shepmaster:
> * any additional tests, which may or may not actually be a deal breaker for merging

I got interested to play with the new `.await` syntax, so I started poking at a potential implementation for that. Writing the tests showed that it would actually be very painful to use without extra turbofishes. For example, this code works fine:

```rust
async fn load_stock_data_sequential() -> Result<String, Error> {
    let apple = api::fetch_page("apple").await.context(UnableToLoadAppleStock)?;
    let google = api::fetch_page("google").await.context(UnableToLoadGoogleStock)?;
    Ok(format!("{}+{}", apple, google))
}
```

But this would not:

```rust
async fn load_stock_data_concurrent() -> Result<String, Error> {
    let apple = api::fetch_page("apple").context(UnableToLoadAppleStock);
    let google = api::fetch_page("google").context(UnableToLoadGoogleStock);

    let (apple, google) = future::try_join(apple, google).await?;

    Ok(format!("{}+{}", apple, google))
}
```

It actually requires something like

```rust
let apple = api::fetch_page("apple").context::<_, Error>(UnableToLoadAppleStock);
let google = api::fetch_page("google").context::<_, Error>(UnableToLoadGoogleStock);
```

Which kind of blows away the ergonomic benefit.

I think one solution would be to introduce a trait that could be implemented on the context selector that would provide a unique `Error` type, removing the ambiguity:

```rust
pub trait IntoError {
    type SourceError; // std::error::Error ?
    type Error: std::error::Error; // ?

    fn into_error(self, error: Self::SourceError) -> Self::Error;
}
```

Quickly scratching out the idea seems to show that this could work. 

This *might* also touch on @BurntSushi's request to have a way to create errors more directly, as it is called like:

```rust
SomeContext { /* ... */ }.into_error(error)
```

shepmaster:
If you'd like to see my WIP, it's [in the branch std-futures](https://github.com/shepmaster/snafu/tree/std-futures).

src/lib.rs
shepmaster: Using the proposed `IntoError` trait actually removes these requirements completely.

src/lib.rs
shepmaster: > something like `futures_crate`

Perhaps just `futures01` and then the standard library can be `futures`.

src/lib.rs
shepmaster: This should be something like "cannot poll ContextFuture after it has resolved", as it's totally fine to poll it twice before that.

shepmaster:
We should also follow the pattern from the futures crate and use `Fut` as the generic type for a wrapped future.

shepmaster:
I was wrong about being able to completely eliminate the `PhantomData`. Since we allow for generic error types, there will always be a need for something to propagate the type chosen from `.context` through to the combinator.

I've updated my branch of standard-compatible futures to account for this.

shepmaster:
@ErichDonGubler will you have any time in the next few days to take this over the finish line, or would you like me to pick it up where it is and add the final polish?

ErichDonGubler:
It would be wonderful if you could keep going with it! :) I've been meaning to write as much all day here, and I appreciate the ping.

shepmaster:
@ErichDonGubler I've laid some groundwork in other PRs and then applied the feedback here. Care to give this a look-over?

shepmaster:
I'm going to go ahead and merge this; thanks!

PR #83, opened 2019-05-18
No longer implement `Borrow`

### LINKED ISSUE #75: Investigate alternatives to using `Borrow`
The root problem boils down to the implementation of `Error::source`. This code works, but the match arms are heterogeneous and cannot be automatically determined / generated via a procedural macro:

```rust
use std::{error, fmt, io};

#[derive(Debug)]
enum Error {
    TraitObjectSendSync(Box<dyn error::Error + Send + Sync + 'static>),
    TraitObjectSend(Box<dyn error::Error + Send + 'static>),
    TraitObjectSync(Box<dyn error::Error + Sync + 'static>),
    TraitObject(Box<dyn error::Error + 'static>),
    Boxed(Box<io::Error>),
    Unboxed(io::Error),
}

impl fmt::Display for Error {
    fn fmt(&self, _: &mut fmt::Formatter) -> fmt::Result {
        Ok(())
    }
}

impl error::Error for Error {
    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
        use Error::*;

        Some(match *self {
            TraitObjectSendSync(ref e) => &**e,
            TraitObjectSend(ref e) => &**e,
            TraitObjectSync(ref e) => &**e,
            TraitObject(ref e) => &**e,
            Boxed(ref e) => e,
            Unboxed(ref e) => e,
        })
    }
}
```

The "safest" solution is to require that users annotate if the source is a trait object:

```rust
#[derive(Snafu)]
enum Error {
    Boom { 
        #[snafu(source(trait_object))]
        source: Box<dyn std::error::Error>,
    },
}
```

This sucks from an ergonomics perspective.

We could do some heuristics of "does the type contain `Box`" or "does the type contain `dyn`", but any heuristic can fail and we'd need a way to override them. Maybe it's `Rc` and not `Box` or the user doesn't use `dyn`, or they have a type alias for the trait object.

See also:
- #74 
- #59 
- f805fc9057af958b3fc49fb2e060ecc908bfdc41
- [Zulip 1](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/.26Box.3Cdyn.20Error.3E.20can't.20be.20.26dyn.20Error)
- [Zulip 2](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Homogenous.20conversion.20of.20Error.20types.20and.20trait.20objects)

shepmaster:
Fixes #75

### DIFF
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -645,14 +645,12 @@ trait GenericAwareNames {
 impl EnumInfo {
     fn generate_snafu(self) -> proc_macro2::TokenStream {
         let context_selectors = ContextSelectors(&self);
-        let borrow_impl = BorrowImpl(&self);
         let display_impl = DisplayImpl(&self);
         let error_impl = ErrorImpl(&self);
         let error_compat_impl = ErrorCompatImpl(&self);
 
         quote! {
             #context_selectors
-            #borrow_impl
             #display_impl
             #error_impl
             #error_compat_impl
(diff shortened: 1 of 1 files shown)


tjkirch:
Does this mean the Borrow reference in the guide is now incorrect, or were there different implementations of Borrow?  Sorry, just want to make sure I understand.

https://docs.rs/snafu/0.4.1/snafu/guide/the_macro/index.html#borrow

shepmaster:
> the Borrow reference in the guide is now incorrect

That's annoying! Yes, that section should be removed now.

PR #107, opened 2019-07-07
Implement Doc Comments as Default Display Implementations

### LINKED ISSUE #106: Use doc comments as default Display implementations
A lot of my errors look like this with snafu:
```rust
/// The Error type for splits files that couldn't be parsed by the FaceSplit
/// Parser.
#[derive(Debug, snafu::Snafu)]
pub enum ParseError {
    /// Expected the title, but didn't find it.
    #[snafu(display("Expected the title, but didn't find it."))]
    ExpectedTitle,
    /// Expected the goal, but didn't find it.
    #[snafu(display("Expected the goal, but didn't find it."))]
    ExpectedGoal,
    /// Expected the attempt count, but didn't find it.
    #[snafu(display("Expected the attempt count, but didn't find it."))]
    ExpectedAttemptCount,
    /// Expected the name of the segment, but didn't find it.
    #[snafu(display("Expected the name of the segment, but didn't find it."))]
    ExpectedSegmentName,
    /// Expected the split time, but didn't find it.
    #[snafu(display("Expected the split time, but didn't find it."))]
    ExpectedSplitTime,
    /// Expected the best segment time, but didn't find it.
    #[snafu(display("Expected the best segment time, but didn't find it."))]
    ExpectedBestSegmentTime,
}
```
As you can see all the display attributes are basically verbatim copied into the doc comments. It would be cool if Snafu used the doc comment as the display impl for a variant if there is no explicit snafu(display(...)) attribute.

CryZe:
If there is no #[snafu(display(...))] attribute on a variant, but there is a doc comment, the doc comment is now chosen as the implementation of Display for the variant.

Resolves #106

### DIFF
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -44,6 +44,7 @@ struct VariantInfo {
     backtrace_delegate: Option<Field>,
     user_fields: Vec<Field>,
     display_format: Option<UserInput>,
+    doc_comment: String,
     visibility: Option<UserInput>,
 }
 
@@ -157,6 +158,8 @@ fn parse_snafu_enum(
 
             let mut display_format = None;
             let mut visibility = None;
+            let mut doc_comment = String::new();
+            let mut reached_end_of_doc_comment = false;
 
             for attr in attributes_from_syn(variant.attrs)? {
                 match attr {
@@ -164,6 +167,22 @@ fn parse_snafu_enum(
                     SnafuAttribute::Visibility(v) => visibility = Some(v),
                     SnafuAttribute::Source(..) => { /* Report this isn't valid here? */ }
                     SnafuAttribute::Backtrace(..) => { /* Report this isn't valid here? */ }
+                    SnafuAttribute::DocComment(doc_comment_line) => {
+                        // We join all the doc comment attributes with a space,
+                        // but end once the summary of the doc comment is
+                        // complete, which is indicated by an empty line.
+                        if !reached_end_of_doc_comment {
+                            let trimmed = doc_comment_line.trim();
+                            if trimmed.is_empty() {
+                                reached_end_of_doc_comment = true;
+                            } else {
+                                if !doc_comment.is_empty() {
+                                    doc_comment.push_str(" ");
+                                }
+                                doc_comment.push_str(trimmed);
+                            }
+                        }
+                    }
                 }
             }
 
@@ -214,6 +233,7 @@ fn parse_snafu_enum(
                         },
                         SnafuAttribute::Visibility(_) => { /* Report this isn't valid here? */ }
                         SnafuAttribute::Display(_) => { /* Report this isn't valid here? */ }
+                        SnafuAttribute::DocComment(_) => { /* Just a regular doc comment. */ }
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


CryZe:
Not sure what this CI failure is.

CryZe:
Nvm, looks like more recent rustc simply has slightly different error output.

tests/doc_comment.rs
```diff
@@ -0,0 +1,16 @@
+extern crate snafu;
+
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+enum Error {
+    /// No user available
```
shepmaster: How does this look for a pathological docstring, such as

```rust
/// No user available. This is unlikely to occur.
///
/// Call me Ishmael, and lo the slithy toves did lorem ipsum…
```

Should we detect when two consecutive newlines have been provided and only take values before that?

snafu-derive/src/lib.rs
```diff
@@ -569,19 +581,31 @@ impl syn::parse::Parse for SnafuAttributeBody {
...
     let mut errs = Vec::new();
 
-    let parsed_attrs = attrs
-        .into_iter()
-        .filter(|attr| attr.path.is_ident("snafu"))
-        .map(|attr| {
-            let body: SnafuAttributeBody = parse2(attr.tts)?;
-            Ok(body.0)
-        });
+    let parsed_attrs = attrs.into_iter().filter_map(|attr| {
```
shepmaster: Could we change this to `flat_map`? I tend to prefer it as it’s strictly more powerful than `filter_map`.

snafu-derive/src/lib.rs
```diff
@@ -569,19 +581,31 @@ impl syn::parse::Parse for SnafuAttributeBody {
     }
 }
 
+struct DocComment(SnafuAttribute);
+
+impl syn::parse::Parse for DocComment {
+    fn parse(input: syn::parse::ParseStream) -> SynResult<Self> {
+        let _: Token![=] = input.parse()?;
```
shepmaster: It’s silly, but I’m avoiding using `Token!` macro in favor of the longer-form `use syn::token::…`;

shepmaster:
I’m also pulling out the compile-fail commit to a new PR (#108) to get it merged early. There’s some other people working on stuff, so they might hit it. You’ll want to rebase once it’s merged.

tests/doc_comment.rs
CryZe: Yeah sounds like a good idea. I implemented this.

CryZe:
This is done btw :)

tests/doc_comment.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+enum Error {
+    /// No user available.
+    /// You may need to specify one.
+    ///
+    /// Here is a more detailed description.
+    MissingUser,
```
shepmaster: Can we add one more variant here that shows that `#[snafu(display)]` overrides a doc comment?

CryZe:
Done :)

shepmaster:
Thanks very much!

PR #109, opened 2019-07-08
Improve errors for misused attributes

### LINKED ISSUE #105: Error when attributes have been misused
Find and fix all the [`Report this isn't valid here` comments](https://github.com/shepmaster/snafu/blob/52e16704228abb9310f74ba74f1b755383c17ada/snafu-derive/src/lib.rs#L165). Super annoying!

tjkirch:
Fixes #105.

The first commit directly addresses the discussion in #105 by adding errors for cases of attribute misuse that were known to be wrong but didn't yet have errors.  Previously, the derive would just continue without the effect the user was trying to request.

The second commit addresses similar comments elsewhere in the file that suggested we warn if attributes are duplicated.  This wasn't discussed in #105 but seemed closely related.

---

**Testing done:**

Existing unit/doc tests still pass, and I added compile-fail tests to exercise each of the new error cases.

### DIFF
--- a/compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
+++ b/compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
@@ -1,0 +1,51 @@
+mod enum_duplication {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub))]
+    #[snafu(visibility(pub))]
+    enum EnumError {
+        #[snafu(display("an error variant"))]
+        #[snafu(display("should not allow duplicate display"))]
+        #[snafu(visibility(pub))]
+        #[snafu(visibility(pub))]
+        AVariant {
+            #[snafu(source(from(EnumError, Box::new)))]
+            #[snafu(backtrace(delegate))]
+            source: Box<EnumError>,
+            #[snafu(source(from(EnumError, Box::new)))]
+            #[snafu(backtrace(delegate))]
+            source2: Box<EnumError>,
+            #[snafu(source)]
+            #[snafu(backtrace(delegate))]
+            source3: String,
+            #[snafu(source)]
+            #[snafu(source)]
+            #[snafu(backtrace(delegate))]
+            source4: String,
+
+            #[snafu(backtrace)]
+            backtrace1: String,
+            #[snafu(backtrace)]
+            backtrace2: String,
+            #[snafu(backtrace)]
+            #[snafu(backtrace)]
+            #[snafu(backtrace)]
+            backtrace3: String,
+        },
+    }
+}
+
+mod struct_duplication {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    enum UsableError {}
+
+    #[derive(Debug, Snafu)]
+    #[snafu(source(from(UsableError, Box::new)))]
+    #[snafu(source(from(UsableError, Box::new)))]
+    struct StructError(Box<UsableError>);
+}
+
+fn main() {}
--- a/compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
+++ b/compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
@@ -1,0 +1,83 @@
+error: Multiple 'visibility' attributes are not supported on an error enum; found a duplicate here:
+ --> $DIR/attribute-duplication.rs:7:10
+  |
+7 |     enum EnumError {
+  |          ^^^^^^^^^
+
+error: Multiple 'source' attributes are not supported on one field; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:25:13
+   |
+25 |             source4: String,
+   |             ^^^^^^^
+
+error: Multiple 'backtrace' attributes are not supported on one field; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:34:13
+   |
+34 |             backtrace3: String,
+   |             ^^^^^^^^^^
+
+error: Multiple 'source' attributes are not supported within an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:21:13
+   |
+21 |             source3: String,
+   |             ^^^^^^^
+
+error: Multiple 'source' attributes are not supported within an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:25:13
+   |
+25 |             source4: String,
+   |             ^^^^^^^
+
+error: Multiple 'backtrace' attributes are not supported within an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:30:13
+   |
+30 |             backtrace2: String,
+   |             ^^^^^^^^^^
+
+error: Multiple 'backtrace' attributes are not supported within an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:34:13
+   |
+34 |             backtrace3: String,
+   |             ^^^^^^^^^^
+
+error: Multiple 'backtrace(delegate)' attributes are not supported within an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:21:13
+   |
+21 |             source3: String,
+   |             ^^^^^^^
+
+error: Multiple 'backtrace(delegate)' attributes are not supported within an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:25:13
+   |
+25 |             source4: String,
+   |             ^^^^^^^
+
+error: Cannot have 'backtrace' and 'backtrace(delegate)' fields in the same error variant
+  --> $DIR/attribute-duplication.rs:28:13
+   |
+28 |             backtrace1: String,
+   |             ^^^^^^^^^^
+
+error: Cannot have 'backtrace' and 'backtrace(delegate)' fields in the same error variant
+  --> $DIR/attribute-duplication.rs:15:13
+   |
+15 |             source: Box<EnumError>,
+   |             ^^^^^^
+
+error: Multiple 'display' attributes are not supported on an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:12:9
+   |
+12 |         AVariant {
+   |         ^^^^^^^^
+
+error: Multiple 'visibility' attributes are not supported on an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:12:9
+   |
+12 |         AVariant {
+   |         ^^^^^^^^
+
+error: Multiple 'source(from)' attributes are not supported on an error struct; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:47:25
+   |
+47 |     #[snafu(source(from(UsableError, Box::new)))]
+   |                         ^^^^^^^^^^^
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -110,12 +110,23 @@ struct SyntaxErrors {
 impl SyntaxErrors {
     /// Adds a new syntax error.  The given description will be used in the compile error pointing
     /// to the given span.  Helper structs are available to format common descriptions, e.g.
-    /// OnlyValidOn.
+    /// OnlyValidOn and DuplicateAttribute.
     fn add<D>(&mut self, span: proc_macro2::Span, description: D)
     where
         D: fmt::Display,
     {
         self.inner.push(syn::Error::new(span, description));
+    }
+
+    /// Adds the given list of errors.
+    fn extend(&mut self, errors: Vec<syn::Error>) {
+        self.inner.extend(errors);
+    }
+
+    #[allow(dead_code)]
+    /// Returns the number of errors that have been added.
+    fn len(&self) -> usize {
+        self.inner.len()
     }
 
     /// Consume the SyntaxErrors, returning Ok if there were no syntax errors added, or Err(list)
@@ -169,6 +180,92 @@ impl fmt::Display for IncompatibleAttributes {
     }
 }
 
+/// Helper structure to simplify parameters to SyntaxErrors.add, handling cases where an attribute
+/// was incorrectly used multiple times on the same element.
+#[derive(Debug)]
+struct DuplicateAttribute {
+    attribute: &'static str,
+    location: &'static str,
+}
+
+impl fmt::Display for DuplicateAttribute {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "Multiple '{}' attributes are not supported {}; found a duplicate here:",
+            self.attribute, self.location,
+        )
+    }
+}
+
+/// AtMostOne is a helper to track attributes seen during parsing.  If more than one item is added,
+/// it's added to a list of DuplicateAttribute errors, using the given `name` and `location` as
+/// descriptors.
+///
+/// When done parsing a structure, call `finish`; it gives you the first attribute found, if any,
+/// and the list of errors.
+#[derive(Debug)]
+struct AtMostOne<T> {
+    name: &'static str,
+    location: &'static str,
+    value: Option<T>,
+    errors: SyntaxErrors,
+}
+
+impl<T> AtMostOne<T> {
+    /// Creates an AtMostOne to track an attribute with the given `name` on the given `location`.
+    /// `location` is a string describing where the attribute can be specified, often referencing a
+    /// parent element, for example "on a field".
+    fn new(name: &'static str, location: &'static str) -> Self {
+        Self {
+            name,
+            location,
+            value: None,
+            errors: SyntaxErrors::default(),
+        }
+    }
+
+    /// Add an occurence of the attribute found at the given `span`.
+    fn add(&mut self, item: T, span: proc_macro2::Span) {
+        if self.value.is_some() {
+            self.errors.add(
+                span,
+                DuplicateAttribute {
+                    attribute: self.name,
+                    location: self.location,
+                },
+            );
+        } else {
+            self.value = Some(item);
+        }
+    }
+
+    #[allow(dead_code)]
+    /// Returns the number of elements that have been added.
+    fn len(&self) -> usize {
+        if self.value.is_some() {
+            1 + self.errors.len()
+        } else {
+            0
+        }
+    }
+
+    /// Returns true if no elements have been added, otherwise false.
+    fn is_empty(&self) -> bool {
+        self.value.is_none()
+    }
+
+    /// Consumes the AtMostOne, returning the first item added, if any, and the list of errors
+    /// representing any items added beyond the first.
+    fn finish(self) -> (Option<T>, Vec<syn::Error>) {
+        let errors = match self.errors.finish() {
+            Ok(()) => Vec::new(),
+            Err(vec) => vec,
+        };
+        (self.value, errors)
+    }
+}
+
 fn impl_snafu_macro(ty: syn::DeriveInput) -> TokenStream {
     match parse_snafu_information(ty) {
         Ok(info) => info.into(),
@@ -217,12 +314,12 @@ fn parse_snafu_enum(
 
     let mut errors = SyntaxErrors::default();
 
-    let mut default_visibility = private_visibility();
+    let mut default_visibilities = AtMostOne::new("visibility", "on an error enum");
 
     for attr in attributes_from_syn(attrs)? {
         match attr {
             SnafuAttribute::Visibility(v) => {
-                default_visibility = v;
+                default_visibilities.add(v, name.span());
             }
             SnafuAttribute::Display(..) => {
                 errors.add(
@@ -274,21 +371,25 @@ fn parse_snafu_enum(
         }
     }
 
+    let (maybe_default_visibility, errs) = default_visibilities.finish();
+    let default_visibility = maybe_default_visibility.unwrap_or_else(private_visibility);
+    errors.extend(errs);
+
     let variants: sponge::AllErrors<_, _> = enum_
         .variants
         .into_iter()
         .map(|variant| {
             let name = variant.ident;
 
-            let mut display_format = None;
-            let mut visibility = None;
+            let mut display_formats = AtMostOne::new("display", "on an error variant");
+            let mut visibilities = AtMostOne::new("visibility", "on an error variant");
             let mut doc_comment = String::new();
             let mut reached_end_of_doc_comment = false;
 
             for attr in attributes_from_syn(variant.attrs)? {
                 match attr {
-                    SnafuAttribute::Display(d) => display_format = Some(d),
-                    SnafuAttribute::Visibility(v) => visibility = Some(v),
+                    SnafuAttribute::Display(d) => display_formats.add(d, name.span()),
+                    SnafuAttribute::Visibility(v) => visibilities.add(v, name.span()),
                     SnafuAttribute::Source(..) => {
                         errors.add(
                             name.span(),
@@ -340,9 +441,10 @@ fn parse_snafu_enum(
             };
 
             let mut user_fields = Vec::new();
-            let mut source_fields = Vec::new();
-            let mut backtrace_fields = Vec::new();
-            let mut backtrace_delegates = Vec::new();
+            let mut source_fields = AtMostOne::new("source", "within an error variant");
+            let mut backtrace_fields = AtMostOne::new("backtrace", "within an error variant");
+            let mut backtrace_delegates =
+                AtMostOne::new("backtrace(delegate)", "within an error variant");
 
             for syn_field in fields {
                 let span = syn_field.span();
@@ -354,10 +456,23 @@ fn parse_snafu_enum(
                     ty: syn_field.ty,
                 };
 
-                let mut has_backtrace = false;
-                let mut is_source = None;
+                let mut has_backtrace_delegate = false;
                 let mut is_backtrace = None;
-                let mut transformation = None;
+                let mut is_source = None;
+                let mut transformations =
+                    AtMostOne::new("source(from)", "within an error variant");
+
+                // Check whether we have multiple source/backtrace attributes on this field.
+                // We can't just add to source_fields/backtrace_fields from inside the attribute
+                // loop because source and backtrace are connected and require a bit of special
+                // logic after the attribute loop.  For example, we need to know whether there's a
+                // source transformation before we record a source field, but it might be on a
+                // later attribute.  We don't actually need any data out of this, it's just a
+                // convenient way to check for dupes and get nice errors.
+                // Note: if there are more than two, the errors generated are identical and dedupe
+                // down to one error; it should be enough to find the problem.
+                let mut source_attrs = AtMostOne::new("source", "on one field");
+                let mut backtrace_attrs = AtMostOne::new("backtrace", "on one field");
 
                 for attr in attributes_from_syn(syn_field.attrs)? {
                     match attr {
@@ -366,7 +481,7 @@ fn parse_snafu_enum(
                             for s in ss {
                                 match s {
                                     Source::Flag(v) => {
-                                        if !v && transformation.is_some() {
+                                        if !v && !transformations.is_empty() {
                                             errors.add(
                                                 name.span(),
                                                 IncompatibleAttributes {
@@ -374,6 +489,9 @@ fn parse_snafu_enum(
                                                     location: "one field",
                                                 },
                                             );
+                                        }
+                                        if v {
+                                            source_attrs.add((), name.span());
                                         }
                                         is_source = Some(v);
                                     }
@@ -387,14 +505,21 @@ fn parse_snafu_enum(
                                                 },
                                             );
                                         }
-                                        transformation = Some((t, e));
+                                        source_attrs.add((), name.span());
+                                        transformations.add((t, e), name.span());
                                     }
                                 }
                             }
                         }
                         SnafuAttribute::Backtrace(b) => match b {
-                            Backtrace::Flag(v) => is_backtrace = Some(v),
-                            Backtrace::Delegate => has_backtrace = true,
+                            Backtrace::Flag(v) => {
+                                backtrace_attrs.add((), name.span());
+                                is_backtrace = Some(v);
+                            },
+                            Backtrace::Delegate => {
+                                backtrace_attrs.add((), name.span());
+                                has_backtrace_delegate = true;
+                            }
                         },
                         SnafuAttribute::Visibility(_) => {
                             errors.add(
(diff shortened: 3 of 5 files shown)


shepmaster:
> The invalid cases can't all be tested together, because the struct derive depends on the enum being valid

Yep, there’s “phases” of compilation where we can’t continue. My hope is basically to make those phases as few and large as possible, but there will always be *some* case where it just doesn’t work.

> * throwing warnings, but it seemed better to me to throw errors

I don’t know that there’s an actual way to generate warnings, either…

> * Should there be unit tests showing that these new cases throw errors

I’m sorry that I forgot to mention this! Take a look at the [compile-fail tests](https://github.com/shepmaster/snafu/tree/master/compatibility-tests/compile-fail/tests/ui). These assert that the error messages continue to work, and I strongly agree that we should have tests for each of the new errors.

I don’t know yet if “more errors in one file” or “more files” is a better approach. Probably more files, but I have a feeling that will be slower in CI at some point. For now, I’d suggest creating at least one new file for these errors.

You’ll want to rebase on master to get the fixes to these tests (causing your CI failure)

> * are there any concerns about existing use cases breaking because we throw new errors

This is where warnings would be nice. They could be warnings for the current semver release, and made into errors for the next breaking change. 

Since we are still pre-1.0, I don’t mind bumping the version early and often, so I’d probably move to 0.5 after this is merged if we can’t figure out warnings. If it was already 1.0, I’d be more torn about a major bump for such a thing, as it was code that wasn’t working the way people intended anyway.

tjkirch:
> compile-fail tests

Ah, thanks!  Hopefully I can add these tonight.

> probably move to 0.5 after this is merged

Would it help if I add commits for the update?  It looks like it'd be `snafu-derive/Cargo.toml`, `Cargo.toml`, `CHANGELOG.md`, and `src/guide/upgrading.md`.  I'm not sure what else you'd like in `CHANGELOG` but I could start it...

Or, that could wait until we have more research on warning capabilities.

shepmaster:
> if I add commits for the update

Thank you, but that’s all right! I’m lazy and don’t keep the changelog up-to-date with each PR, only doing it when I do a release. It does provide a good opportunity to review the changes and make sure that nothing obvious is missing or untested.

> `CHANGELOG.md` and `src/guide/upgrading.md`

Feel free to suggest some text that you’d like to see as a seed for what will end up here!

tjkirch:
Thanks for all your guidance @shepmaster!  I marked this as ready for review, since I think I've addressed your concerns, at least for an initial implementation.

compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
```diff
@@ -0,0 +1,28 @@
+mod enum_duplication {
```
shepmaster: Did you also want a module for structs? Otherwise, I think it’s ok to remove the module here.

snafu-derive/src/lib.rs
```diff
@@ -165,8 +167,18 @@ fn parse_snafu_enum(
                 match attr {
                     SnafuAttribute::Display(d) => display_format = Some(d),
                     SnafuAttribute::Visibility(v) => visibility = Some(v),
-                    SnafuAttribute::Source(..) => { /* Report this isn't valid here? */ }
-                    SnafuAttribute::Backtrace(..) => { /* Report this isn't valid here? */ }
+                    SnafuAttribute::Source(..) => {
+                        errors.push(syn::Error::new(
+                            name.span(),
+                            "'source' attribute is only valid on individual fields; attempted to place on this variant:",
+                        ));
```
shepmaster: I wonder if there’s anything we can do about the “programmer overhead” of creating these, as well as the duplicated phrases. Potentially, this could look like:

```rust
errors.only_on_field(name, “source”)
errors.only_on_field(name, “backtrace”)
```

As a minimum, this could be used to hide the inner `Vec` and construction of `Error`:

```rust
impl MyErrorCollectionName {
    fn add(&mut self, span: impl Spanned, text: impl Into<String>) { /* … */ }

    fn finish(self) -> Result<(), Vec<…>> { /* … */ }
}
```

snafu-derive/src/lib.rs
```diff
@@ -262,15 +284,46 @@ fn parse_snafu_enum(
                 }
             }
 
+            // Reverse the lists of fields where we may find duplicates so that we can call the
+            // second (and later) items the "duplicates" without uglier indexing.
```
shepmaster: I like it 😁

snafu-derive/src/lib.rs
```diff
@@ -262,15 +284,46 @@ fn parse_snafu_enum(
...
-            // Report a warning if there are multiple?
-            // Report a warning if delegating and our own?
+            for extra_field in backtrace_delegates.into_iter() {
+                errors.push(syn::Error::new(
+                    extra_field.name.span(),
+                    "Only one field can be designated as 'backtrace(delegate)'; found a duplicate here:",
+                ));
+            }
+
+            if backtrace_field.is_some() && backtrace_delegate.is_some() {
```
shepmaster: Can this be something like `if let Some(backtrace_field,  backtrace_delegate) = (backtrace_field.as_ref(), backtrace_delegate.as_ref())` to avoid the `unwrap`s?

snafu-derive/src/lib.rs
```diff
@@ -305,19 +363,41 @@ fn parse_snafu_struct(
...
+                ));
+            }
             SnafuAttribute::Source(ss) => {
                 for s in ss {
                     match s {
-                        Source::Flag(..) => { /* Report this isn't valid here? */ }
+                        Source::Flag(..) => {
+                            errors.push(syn::Error::new(
+                                name.span(),
+                                "'source(bool)' is only valid on individual fields; attempted to place on this struct:",
```
shepmaster: This one sticks out because it’s the only one that says `(bool)` — why the difference?

tjkirch:
> There’s one place where I was lazy and didn’t put the “report this” text:
> 
> https://github.com/shepmaster/snafu/blob/da2d4aced258bdc443c90d8b3de6b127912e2af0/snafu-derive/src/lib.rs#L147-L151
> 
> Can you expand that code to also report the attributes?

To make sure I understand this - you mean changing from this functional style to a loop that checks all of the attributes from `attributes_from_syn` and adds errors for anything that's not `visibility`?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
tjkirch: I think there should be a module for structs, but I didn't get struct duplication checks working - the code structure for structs wasn't laid out the way it was for enums, and I was having a harder time figuring out how to do that.  I wanted to leave this module so it'd be clearer that enums aren't the only things to check.

shepmaster:
> changing from this functional style to a loop

Yep. I don’t see an obvious way, but if you see something that’s a bit more functional-style, feel free to go for it!

snafu-derive/src/lib.rs
tjkirch: I lean toward the latter because I think having a custom message can be helpful for clarifying what the user actually did that wasn't quite right.  I'll give that a shot.

snafu-derive/src/lib.rs
tjkirch: Ah, I didn't know that syntax would work for combining `Option`s - I'll give it a shot.

snafu-derive/src/lib.rs
tjkirch: This one was confusing to me.  I was trying to figure out what `Flag` is, and I think it's when you use `source(false)` to say a field isn't the source even though it's named `source`.  You'd just say `source` if it *is* the source, so `source(true)` shouldn't really be necessary.  That's my understanding of this code path, and I used `source(bool)` as a shorthand for it, since I believe it would only happen if the user did specify a boolean inside `source()`.  Is that right?

snafu-derive/src/lib.rs
shepmaster: > I lean toward the latter

I don’t think they have to be either-or:

```rust
impl MyErrorCollectionName {
    fn add_only_on_field(&mut self, span: impl Spanned, field: &str) {
        self.add(span, format!(“This is repeated text for field {}, do this other thing instead”));
    }

    fn add(&mut self, span: impl Spanned, text: impl Into<String>) { /* … */ }
}
```

tjkirch: Ahh, I see what you mean.  "field" works well for handling duplication, but we might also want something like "element type" for misuse?  I think I'd typically want to say something like "this doesn't apply to [field, variant, enum, struct], it's only valid on [field, variant, enum, struct]".  I guess that could just be handled by the custom message, but it's even more common than repetition, so maybe it's worth a try simplifying it too.

shepmaster: Yeah, there's lots of ways this could go. My suggestion would be to start with adding the plain `add` method and then see what repetition remains. That might then highlight new methods to be created. How those look could be as simple as taking in a `&str`, up to taking in a slice of items, or as far as some custom type(s). I think it's worth staying practical with respect to the amount of code we are saving, so the latter option seems unlikely :-) 

As a side note, you can't actually use the `impl ...` syntax I listed because we support Rust 1.18, but hopefully the idea is communicated.

snafu-derive/src/lib.rs
shepmaster: Ah, yes, I see. Your analysis is correct. I was missing that `source(from(...))` was valid here, but `source(false)` wasn't.

> so `source(true)` shouldn't really be necessary

I did this for consistency of `true` / `false` but also *potentially* allowing the concept of multiple attributes that subsequently override each other. There's also a nascent idea of allowing multiple `source(from())` attributes, for example. I wouldn't worry about either of those future plans, but just so you are aware of the background.

snafu-derive/src/lib.rs
tjkirch: I started with `add`, moved to a couple custom methods that covered 90% of the cases, and eventually found that a couple custom types were simpler.  While it's slightly more verbose, I think it's rather clear - hope you like it as much as I do :)

snafu-derive/src/lib.rs
```diff
@@ -98,6 +99,76 @@ impl Transformation {
...
+    /// The name of the attribute that was misused.
+    attribute: &'static str,
+    /// A description of where that attribute is valid.
+    valid_on: &'static str,
+    /// A description of where the attribute was incorrectly used.
+    not_on: &'static str,
+}
+
+impl fmt::Display for OnlyValidOn {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
shepmaster: Sadly, `Foo<'_>` isn't valid in older versions of Rust, so we'll have to leave that off for now.

snafu-derive/src/lib.rs
```diff
@@ -98,6 +99,76 @@ impl Transformation {
...
 
+/// SyntaxErrors is a convenience wrapper for a list of syntax errors discovered while parsing
+/// something that derives Snafu.  It makes it easier for developers to add and return syntax
+/// errors while walking through the parse tree.
+struct SyntaxErrors {
+    inner: Vec<syn::Error>,
+}
+
+impl SyntaxErrors {
+    fn new() -> SyntaxErrors {
```
shepmaster: This is fine, but for what it's worth, I would have just added `#[derive(Default)]` and then called `SyntaxErrors::default()`

snafu-derive/src/lib.rs
shepmaster: > eventually found that a couple custom types were simpler

I don't disagree, but would you mind expanding on bit on how you found the methods lacking?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
shepmaster: > I didn't get struct duplication checks working

Can you expand a bit on what problems arose?

tjkirch: It was just that the enum-handling code's structure made it obvious what to do, and the struct-handling code wasn't as obvious for me, and I wasn't comfortable enough yet with the code style.

Now that I look back at it, I think there may only be one line to change, where we match `Source::From(t, e)` -- if we already have `Some`, then instead of setting `transformation`, we'd add an error.  Does that seem right?

snafu-derive/src/lib.rs
tjkirch: Sure - in particular it was `SyntaxErrors.only_valid_on`, which became the `OnlyValidOn` struct.  It had 3 arguments for descriptive strings, and 1 for the span.  The 3 string arguments were too easy for a developer to mix up, in my opinion -- you always had to check the docs.  With the struct, where you specify its fields at creation, it becomes obvious, and I think it reads very nicely.

(I think having those 3 fields are important to clarify the different types of errors, and clearly point out to the user what they've done wrong.)

snafu-derive/src/lib.rs
tjkirch: Sorry - sure, that's trivial.  I had just copied the `fn fmt` declaration from my current Rust source :)

snafu-derive/src/lib.rs
tjkirch: I'm not sure why I usually forget that `Default` is a thing, but this is an easy improvement I'll make. :)

compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
tjkirch: (And if that's the case, the same thing applies to enums; it's another type of duplication I hadn't caught.)

snafu-derive/src/lib.rs
shepmaster: > The 3 string arguments were too easy for a developer to mix up, in my opinion

I had a hunch this was it. For future reference, one thing I've seen Niko Matsakis do is pseudo-named-arguments:

```rust
struct Attribute(&'static str);
struct ValidOn(&'static str);
struct NotOn(&'static str);

fn only_valid_on(attribute: Attribute, valid_on: ValidOn, not_on: NotOn) {
    println!(
        "'{}' attribute is only valid on {}; attempted to place on this {}:",
        attribute.0, valid_on.0, not_on.0
    )
}

fn main() {
    only_valid_on(Attribute("a"), ValidOn("b"), NotOn("c"));
}
```

compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
shepmaster: > if we already have `Some`, then instead of setting `transformation`, we'd add an error

That would be fine, but we have [similar cases elsewhere](https://github.com/shepmaster/snafu/blob/da2d4aced258bdc443c90d8b3de6b127912e2af0/snafu-derive/src/lib.rs#L415), where we "pop one and exactly one" value from a `Vec`. 

Could we change these cases to push to a `Vec` and then reuse / repurpose that logic:

```rust
fn only_one<T>(v: Vec<T>) -> Option<T> { ... }

let from = only_one(froms).ok_or_else(...)
```

shepmaster: Not 100% sure how to combine that with cases of zero-or-one, or as a not-immediately-fatal error.

tjkirch: I'll give this a shot.

snafu-derive/src/lib.rs
tjkirch: I'm happy to go with whichever style you prefer :)

compatibility-tests/compile-fail/tests/ui/attribute-duplication.rs
tjkirch: I came up with something I'm pretty happy with - `AtMostOne`, which replaces those `Vec`s and creates errors for attributes beyond the first.  It handles the struct case and cleans up the enum cases, removes the weird indexing reverse, and builds nicely on `SyntaxErrors`.  I hope you like it.

It didn't replace `one_field_error` which wasn't as good a fit, but if you have ideas for merging the two I could do that.  (I imagine it'd take a `Fn` for the error generation, or something similar, but it seemed like more overhead than it was worth to replace this one case, when the simpler thing replaced all the rest.)

snafu-derive/src/lib.rs
```diff
@@ -347,12 +627,15 @@ fn parse_snafu_struct(
...
-        Some((ty, expr)) => Transformation::Transform { ty, expr },
-        None => Transformation::None {
+    let (maybe_transformation, errs) = transformations.finish();
+    let transformation = maybe_transformation
+        .map(|(ty, expr)| Transformation::Transform { ty, expr })
+        .unwrap_or(Transformation::None {
             ty: inner.into_value().ty,
-        },
-    };
+        });
```
shepmaster: FYI, [`unwrap_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else) and [`map_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else) exist.

This currently unconditionally calls `into_value`, which is a *slight* inefficiency. I'm not really worried about it here or in the earlier case of it.

tjkirch:
> This all seems great; thanks so much for putting up with all my pedantry!

Your knowledge and guidance are invaluable to the community.  I'm thankful to have been able to take advantage, while helping a project I love.  It inspires me to find other ways to contribute.  So, thank you!

> Is there anything else you think we need to add or discuss before merging?

Just the `or_else` fix that I'll make shortly.  I agree with you that the spans the new errors point to aren't ideal, but it would require a bit more wide-reaching changes that aren't strictly related, so I think it should be a separate issue.  I do think this change sets it up for success with how it accepts/stores spans, but the right ones need to be made available.  Do you agree that should be another issue?  I can make it if so.

snafu-derive/src/lib.rs
tjkirch: Oops; that line was a copy/paste, and I missed the change in behavior in this new context.  Will fix shortly.

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
```diff
@@ -0,0 +1,71 @@
+error: Only one field can be designated as 'visibility'; found a duplicate here:
```
shepmaster: This isn't a field, is it? Shouldn't this say something like "multiple visibility attributes are not supported on an error enum"?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
```diff
@@ -0,0 +1,71 @@
...
+ --> $DIR/attribute-duplication.rs:7:10
+  |
+7 |     enum EnumError {
+  |          ^^^^^^^^^
+
+error: Only one field can be designated as 'source'; found a duplicate here:
```
shepmaster: How about we add "per variant" to this class of error?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
```diff
@@ -0,0 +1,71 @@
...
+18 |             source2: String,
+   |             ^^^^^^^
+
+error: Only one field can be designated as 'backtrace(delegate)'; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:21:13
+   |
+21 |             source3: String,
+   |             ^^^^^^^
+
+error: Cannot have 'backtrace' and 'backtrace(delegate)' fields in the same enum
```
shepmaster: In the same enum or in the same variant?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
```diff
@@ -0,0 +1,71 @@
...
+24 |             backtrace1: String,
+   |             ^^^^^^^^^^
+
+error: Cannot have 'backtrace' and 'backtrace(delegate)' fields in the same enum
+  --> $DIR/attribute-duplication.rs:15:13
+   |
+15 |             source: Box<EnumError>,
+   |             ^^^^^^
+
+error: Only one field can be designated as 'display'; found a duplicate here:
```
shepmaster: ditto: not a field, but duplicate attrs per variant

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
```diff
@@ -0,0 +1,71 @@
...
+  --> $DIR/attribute-duplication.rs:12:9
+   |
+12 |         AVariant {
+   |         ^^^^^^^^
+
+error: Only one field can be designated as 'visibility'; found a duplicate here:
```
shepmaster: ditto: not a field, but duplicate attrs per variant

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
tjkirch: You're right, the context isn't as good as it should be, even without more detailed spans.  I'll work on improving these.

tjkirch:
> I re-read some of the error text and it seems like a few errors got mixed up; am I losing it?

Nope!  I didn't have the right context in a few spots.  I've cleaned those up (easy to do by adding a context `location` to the `AtMostOne`) and tried to make the error wording more consistent with yours.

While doing this, I found a couple more cases of duplication I'm not catching yet, which I think I can fix fairly quickly with the same tools:

* We catch `backtrace*` on multiple fields in a variant, but not multiple `backtrace*` on a single field
* We don't catch `source(from)` and `source(flag)` together on a variant field (only wrong if flag=true?)

shepmaster:
Thanks for the last push! Everything looks good to me now; are there any outstanding issues (other than trying to find better things to point to) we should address before merging and releasing?

tjkirch:
> Thanks for the last push! Everything looks good to me now; are there any outstanding issues (other than trying to find better things to point to) we should address before merging and releasing?

There are a couple more cases of duplication that I mentioned in https://github.com/shepmaster/snafu/pull/109#issuecomment-513607010 -- they're a little harder to fix than I initially expected, but if you're not worried about those, or you'd like them done in a separate change, then no blockers that I know of. :)

shepmaster:
Oh, I assumed that they were part of your last force push. I'm in no hurry, I just don't want to forget about this PR (something I unfortunately have a tendency to do). Please feel free to poke me when you have a chance to cycle back to those last few. Feel free to ask questions if you think it's something I can help with!

tjkirch:
> Feel free to ask questions if you think it's something I can help with!

I'd like to clarify whether some combinations of attributes should be errors, to be sure I'm not going to change behavior.

* I'm guessing that `source(false)` and `source(from(...))` on the same field would be wrong, but I'm not sure - maybe the user is effectively saying "this isn't the source *directly*, but you can make it *from* ...".  Error?
* Would you want to allow `source(true)` and `source(from(...))` on the same field?  Thinking charitably, the user could be double-stating their intention, but it feels more like a confusion about the `from`.  Error?
* I'm guessing `backtrace(false)` and `backtrace(delegate)` on the same field would be wrong, though it's less clear - they could be trying to say "*this* shouldn't backtrace, look inside this delegate".  Still feels like a confusion to me.  Error?
* How about `backtrace(true)` and `backtrace(delegate)`?  That seems even weirder to me.  Error?

Separate from those, I think there are a couple more cases to clean up regardless:
* a field is only intended to have a backtrace delegate if it's a source, but a backtrace delegate on a non-source field is quietly ignored - I think that should be an error.
* a field shouldn't be a non-delegate backtrace and a source, but we quietly ignore the backtrace attribute on a source - I think it should be an error.

As for why it's more challenging than anticipated, aside from the above, it's the interplay between `has_backtrace`, `is_backtrace`, and `is_source` in the field handling code.  I understand what they do now, I'm just trying to find a clean pattern for accomplishing the same checks.  I lean toward adding sources/backtraces to the higher-level AtMostOne as soon as we find them, and using the `is_`/`has_` fields for special error checks (mentioned above) after the attribute loop.  I haven't tried that yet (it'd be nice to confirm the above confusions first) but I'll update this issue if I have trouble.

shepmaster:
> * `source(true)` and `source(from(...))` on the same field
> * `source(false)` and `source(from(...))` on the same field

I could go either way on these. `true` and `from` together don't bother me. I see `true` as meaning "this is a source that cannot be named `source`" and `from` as "here is how you transform from one to the other". It may be better if `from` implied `true`, which it doesn't seem like it does currently.

`false` and `source` do seem like a point of confusion.

> * a non-delegate backtrace and a source

This is interesting. I wonder if we should have made it such that `#[snafu(source, backtrace)]` implicitly meant `delegate`. Then we wouldn't have needed the surface-level API for `backtrace(delegate)` at all. That might also remove some possibilities that we need to prevent.

Do you think that this would be confusing to end users?

> I lean toward adding sources/backtraces to the higher-level AtMostOne as soon as we find them, and using the `is_`/`has_` fields for special error checks (mentioned above) after the attribute loop

Sounds reasonable. I think you'd use the `AtMostOne` before/during the loop, call `finish` (collecting errors), then perform a `match` on the resulting `Option`s to ensure that there's only a reasonable combination.

tjkirch:
> `true` and `from` together don't bother me

That makes sense to me, I'll make sure this still works.

> It may be better if `from` implied `true`

That makes sense and it's easy to do, I'll add that.

> `false` and `source` do seem like a point of confusion.

Ok, I'll make this an error.

> I wonder if we should have made it such that `#[snafu(source, backtrace)]` implicitly meant `delegate`

That's a good idea.  Any `backtrace` on a `source` would be the equivalent of today's `delegate`, and on a non-`source` wouldn't change.  I also lean toward that simpler API.  As a new user, I think it does have a slight potential for confusion, but no more than `delegate` itself, to be honest :)

I think I could do this.  (It'd be a third commit, of course.)

shepmaster:
Sure. If at any point you think that this PR is getting too big, I definitely don't mind merging and creating some new issues / PRs to iterate.

tjkirch:
> If at any point you think that this PR is getting too big, I definitely don't mind merging and creating some new issues / PRs to iterate.

I think the `delegate` change could definitely be separate; my concern there was just that merging these breaking changes, without having the remaining breaking change ready, could impact your ability to cut a release.  If you wanted to or had to for other reasons, you might need yet another release for the breaking `delegate` change.

If you're not worried about that, then all that would remain for this PR is preventing `source(false)` and `source(from(...))` on the same field (as a form of misuse).

shepmaster:
We can merge this but not immediately release anything; bundling up the breaking changes. If we need to release a patch version, I can always fork from before these commits.

tjkirch:
Ok!  I was being perhaps overly cautious.

To recap, here's the remaining work that I think could fit in this PR:
* Error on `source(false)` and `source(from)` on one field  (still allowing `source(true)` and `source(from)`)
* Error on multiple `source` or `backtrace` on one field

These could be other PRs, and I'll make issues for them:
* Making `source(from)` imply `source(true)`
* Removing `backtrace(delegate)` in favor of `backtrace` on a `source` field
* Improving the accuracy of the spans in the new errors

tjkirch:
I've addressed the remaining work for this PR:
* `source(false)` and `source(from)` on a single field now cause an IncompatibleAttributes error
* Multiple `source` or `backtrace` attributes on a single field now error through new AtMostOne trackers

compile-fail tests were added for both.  If you like the approach I took, I think this PR is ready for merging.

With the restructuring, making `source(from)` imply `source(true)` for #124 should be a one-line change in another PR.  It also exposes a couple more duplicates in existing compile-fail tests, because the tests had multiple fields with `source(from)` that weren't counted as sources and therefore not as duplicates :)

tjkirch:
Hmm.  Rust 1.18 fails with a lifetime issue: https://github.com/shepmaster/snafu/pull/109/checks?check_run_id=176857626

I'm not sure offhand what changed between Rust 1.18 and 1.30 that allows this slice to be considered static: https://github.com/shepmaster/snafu/compare/6cb438122453631443eb361840d9499768a9d54a..27504266fd1b266e101b22de103881e84af0a01c#diff-89e1fad4d2637dae0904dce6aba08394R487

@shepmaster - assuming you're OK with this general approach, would you recommend `IncompatibleAttributes` store a `Vec`, and I use a `vec![]` during construction to resolve that lifetime issue, or is there a better way?

tjkirch:
I just learned about `cd compatibility-tests/v1_18 && rustup run 1.18.0 cargo test` so hopefully I can prevent bad pushes in the future - sorry about that.  Another tip I'm thinking about adding to a contributor's guide...

tjkirch:
> I'm not sure offhand what changed between Rust 1.18 and 1.30 that allows this slice to be considered static: https://github.com/shepmaster/snafu/compare/6cb438122453631443eb361840d9499768a9d54a..27504266fd1b266e101b22de103881e84af0a01c#diff-89e1fad4d2637dae0904dce6aba08394R487

I think it might be this change: https://github.com/rust-lang/rust/issues/38865
...but there was also this one that I don't really understand :)  https://github.com/rust-lang/rust/pull/47408

shepmaster:
> making `source(from)` imply `source(true)` for #124 should be a one-line change in another PR.

Nice!

> multiple fields with `source(from)` that weren't counted as sources and therefore not as duplicates

Makes sense 😉 

> what changed between Rust 1.18 and 1.30 that allows this slice to be considered static

Yeah, it'd be the automatic promotion of literals to statics, as you linked.

> or is there a better way

You can do the promotion by hand:

```rust
static FOO: &[&str] = &["source(false)", "source(from)"];
```
```rust
attributes: FOO,
```

> cd compatibility-tests/v1_18 && rustup run 1.18.0 cargo test

There's actually a "rust-toolchain" file in that directory, so you can just do `cd compatibility-tests/v1_18 && cargo test`. If that file wasn't there, I'd recommend doing `cd compatibility-tests/v1_18 && cargo +1.18.0 test` for fewer keystrokes.

> adding to a contributor's guide

That would probably be a good thing!

shepmaster:
> bad pushes in the future - sorry about that

Not at all; that's the point of CI - to catch the things that we forget. That's why I try to be pedantic about adding new tests so frequently. I don't want to have to think about exactly what to remember!

It is useful to check in 1.18 locally only because it's comparatively so different to current Rust that a lot of little things sneak in while developing. My biggest mistake is around match ergonomics.

tjkirch:
> You can do the promotion by hand:

Thanks, that worked!  I just pushed that fix, and hopefully this is everything we need to get this first PR merged.

> There's actually a "rust-toolchain" file in that directory, so you can just do cd compatibility-tests/v1_18 && cargo test. If that file wasn't there, I'd recommend doing cd compatibility-tests/v1_18 && cargo +1.18.0 test for fewer keystrokes.

That's good to know!  The other piece that was missing for me is understanding that `.cirrus.yml` lists some other steps that are required for each of the Rust versions - for example installing precise versions of dependencies that work with that version of Rust.  Without doing anything special, I was hitting errors building dependencies, like `backtrace` using the `crate` keyword in `use` statements.

Do you know if there's a way to run those locally without a lot of copy+paste?  I couldn't find any resources for replicating Cirrus CI locally.  I see they have an issue for making a CLI (https://github.com/cirruslabs/cirrus-ci-docs/issues/108) but didn't see anything else.

shepmaster:
> a way to run those locally without a lot of copy+paste

Nothing I'm aware of. However, it's been pretty standard practice in my other crates to actually have most of the CI-related junk as separate shell scripts so that they are at least easier to access outside of CI (and you get syntax highlighting, etc.)

In this case, I wonder if we should just move those to the respective Cargo.toml files...

shepmaster:
image

What an amazing first contribution! Thanks so much!

tjkirch:
> it's been pretty standard practice in my other crates to actually have most of the CI-related junk as separate shell scripts
>
> In this case, I wonder if we should just move those to the respective Cargo.toml files...

Either of these sound like they would be pretty helpful.  I can create an issue if you like?

> What an amazing first contribution! Thanks so much!

Thanks again for the library, and for your great feedback here!  I hope to keep contributing.

tjkirch:
> Either of these sound like they would be pretty helpful. I can create an issue if you like?

Ah, you already did #128!  Wow!

PR #117, opened 2019-07-18
Allow default generic parameters on error types

### LINKED ISSUE #116: Generic parameters with default values fail because they are placed before generated generic parameters in context selectors
I have a trouble using generic error type with default value, as in snafu-generated context selectors original generic parameters are placed before generated ones and if original generic parameter has default value it triggers compilation error, because defaults must be trailing. Error is very hard to understand as it points to original type where the generic parameter is the only, so is trailing, and you need to know how snafu works to understand what is true cause of that compilation error. I suppose the corresponding source is [https://github.com/shepmaster/snafu/blob/e09e9478bcc37968f711b0d5d22f04ea60963c77/snafu-derive/src/lib.rs#L774](https://github.com/shepmaster/snafu/blob/e09e9478bcc37968f711b0d5d22f04ea60963c77/snafu-derive/src/lib.rs#L774)

fan-tom:
Fixes #116 but I'm not sure about some quote macro usages and test completeness

### DIFF
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -674,6 +674,50 @@ trait GenericAwareNames {
         Box::new(quote! { #enum_name<#(#original_generics,)*> })
     }
 
+    fn provided_generic_types_without_defaults(&self) -> Vec<proc_macro2::TokenStream> {
+        use syn::TypeParam;
+        self.generics()
+            .type_params()
+            .map(|t: &TypeParam| {
+                let &TypeParam {
+                    ref attrs,
+                    ref ident,
+                    ref colon_token,
+                    ref bounds,
+                    ..
+                } = t;
+                quote! {
+                    #(#attrs)*
+                    #ident
+                    #colon_token
+                    #bounds
+                }
+            })
+            .collect()
+    }
+
+    fn provided_generics_without_defaults(&self) -> Vec<proc_macro2::TokenStream> {
+        self.provided_generic_lifetimes()
+            .into_iter()
+            .chain(self.provided_generic_types_without_defaults().into_iter())
+            .collect()
+    }
+
+    fn provided_generic_lifetimes(&self) -> Vec<proc_macro2::TokenStream> {
+        use syn::{GenericParam, LifetimeDef};
+
+        self.generics()
+            .params
+            .iter()
+            .flat_map(|p| match *p {
+                GenericParam::Lifetime(LifetimeDef { ref lifetime, .. }) => {
+                    Some(quote! { #lifetime })
+                }
+                _ => None,
+            })
+            .collect()
+    }
+
     fn provided_generic_names(&self) -> Vec<proc_macro2::TokenStream> {
         use syn::{ConstParam, GenericParam, LifetimeDef, TypeParam};
 
@@ -749,7 +793,10 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
         use syn::Ident;
 
         let enum_name = &self.0.name;
-        let original_generics: &Vec<_> = &self.0.generics.params.iter().collect();
+        let original_lifetimes = self.0.provided_generic_lifetimes();
+        let original_generic_types_without_defaults =
+            self.0.provided_generic_types_without_defaults();
+        let original_generics_without_defaults = self.0.provided_generics_without_defaults();
 
         let parameterized_enum_name = &self.0.parameterized_name();
 
@@ -771,7 +818,7 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
             .as_ref()
             .unwrap_or(&self.0.default_visibility);
 
-        let generics_list = quote! { <#(#original_generics,)* #(#generic_names,)*> };
+        let generics_list = quote! { <#(#original_lifetimes,)* #(#generic_names,)* #(#original_generic_types_without_defaults,)*> };
         let selector_name = quote! { #variant_name<#(#generic_names,)*> };
 
         let names: &Vec<_> = &user_fields.iter().map(|f| f.name.clone()).collect();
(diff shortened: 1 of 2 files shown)


shepmaster:
Thanks for the PR, and sorry for the trouble!

Based on your original report, I would have assumed that the fix would have been to put the user-supplied generic parameters after the automatically-generated ones. Can you share why you add all the other code?

I'll take a look into the CI failure.

fan-tom:
Thanks for reply! If under `all the other code` you mean additional methods in `GenericAwareNames` I see that `provided_generic_types_and_consts` must be deleted(I'll do it tonight), as it is never used. All other changes are done in order to successfully pass existing tests and correctly handle usages of generics with defaults in places other than `impl`, such as `fail`.

snafu-derive/src/lib.rs
```diff
@@ -674,6 +674,57 @@ trait GenericAwareNames {
...
+            .collect()
+    }
+
+    fn provided_generic_lifetimes(&self) -> Vec<proc_macro2::TokenStream> {
+        use syn::{GenericParam, LifetimeDef};
+
+        self.generics()
+            .params
+            .iter()
+            .filter_map(|p| match *p {
```
shepmaster: I prefer to use `flat_map` instead of `filter_map`, since the former is strictly more powerful than the latter. I'd appreciate it if you could change these calls.

snafu-derive/src/lib.rs
```diff
@@ -1121,7 +1184,9 @@ impl<'a> ErrorCompatImpl<'a> {
 
 impl<'a> quote::ToTokens for ErrorCompatImpl<'a> {
     fn to_tokens(&self, stream: &mut proc_macro2::TokenStream) {
-        let original_generics = &self.0.generics;
+        let original_generic_types = &self.0.provided_generic_types_without_defaults();
+        let original_lifetimes = &self.0.provided_generic_lifetimes();
+        let original_generics = original_lifetimes.into_iter().chain(original_generic_types);
```
shepmaster: Since we repeat this pattern 3 times, perhaps it's worth creating a helper method:

```rust
let original_generics = self.0.provided_some_reasonable_name_here();
```

snafu-derive/src/lib.rs
```diff
@@ -749,7 +800,14 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
         use syn::Ident;
 
         let enum_name = &self.0.name;
-        let original_generics: &Vec<_> = &self.0.generics.params.iter().collect();
+        let original_lifetimes: Vec<_> = self.0.generics.lifetimes().collect();
```
shepmaster: How does this differ from the implementation of `provided_generic_lifetimes`?

shepmaster:
> `provided_generic_types_and_consts` must be deleted

I've pushed commits that deny warnings in CI, so hopefully that type of mistake will be harder to make in the future 😉 

> consts

It will be interesting to see how const generics work with SNAFU, but I agree that it's a bit too early to try and add them right now!

> I'll take a look into the CI failure.

This should be fixed; if you rebase on the master branch to pick up the CI warnings changes, the next build should also pick up fixes to pin-project.

fan-tom:
> I haven't pulled this down locally to check the exact behavior, but it looks like the code will take some input with defaults and then discard them for the generated types:
> 
> ```rust
> enum Foo<X = u8>{ /*...*/ }
> 
> // Becomes
> 
> struct FooOne<__T0, X>{ /*...*/ }
> ```
> 
> Is that an accurate analysis of the implementation? Shouldn't we keep the defaults to prevent the user from needing to specify them?

As I can see, currently selectors are generic over only enum field types and doesn't mention original generics at all

```rust
        let generic_names: &Vec<_> = &(0..user_fields.len())
            .map(|i| Ident::new(&format!("__T{}", i), Span::call_site()))
            .collect();

        let visibility = self
            .1
            .visibility
            .as_ref()
            .unwrap_or(&self.0.default_visibility);

        let generics_list = quote! { <#(#original_lifetimes,)* #(#generic_names,)* #(#original_generic_types_without_defaults,)*> };
        let selector_name = quote! { #variant_name<#(#generic_names,)*> };
```

Am I right? Should I also add defaults to generics if they exist?

shepmaster:
Ah, I see what you mean. With the current code...

```rust
extern crate snafu;

use snafu::Snafu;

#[derive(Snafu)]
enum Foo<X = u8>{
    Alpha { my_value: X },
}
```

Expands to (some things removed)

```rust
enum Foo<X = u8> {
    Alpha { my_value: X },
}

struct Alpha<__T0> {
    my_value: __T0,
}

impl<__T0> Alpha<__T0> {
    fn fail<X = u8, __T>(self) -> std::result::Result<__T, Foo<X>>
    where
        __T0: std::convert::Into<X>,
    {
        // ...
    }
}

impl<X = u8, __T0> snafu::IntoError<Foo<X>> for Alpha<__T0>
where
    Foo<X>: std::error::Error + snafu::ErrorCompat,
    __T0: std::convert::Into<X>,
{
    type Source = snafu::NoneError;
    fn into_error(self, error: Self::Source) -> Foo<X> {
        // ...
    }
}

impl<X = u8> std::fmt::Display for Foo<X> {
    // ...
}

impl<X = u8> std::error::Error for Foo<X>
where
    Self: std::fmt::Debug + std::fmt::Display,
{
    // ...
}

impl<X = u8> snafu::ErrorCompat for Foo<X> {
    // ...
}
```

My question was around these pieces:

```rust
impl<__T0> Alpha<__T0> {
    fn fail<X = u8, __T>(self) -> std::result::Result<__T, Foo<X>>
    where
        __T0: std::convert::Into<X>,
    {
        // ...
    }
}

impl<X = u8, __T0> snafu::IntoError<Foo<X>> for Alpha<__T0>
where
    Foo<X>: std::error::Error + snafu::ErrorCompat,
    __T0: std::convert::Into<X>,
{
    type Source = snafu::NoneError;
    fn into_error(self, error: Self::Source) -> Foo<X> {
        // ...
    }
}
```

But I forgot that you *cannot* put the defaults on the `impl` or `fn`:

> defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions.

fan-tom:
Yes, exactly

Issue #113, opened 2019-07-13
Please make "backtraces" feature opt-in, not opt-out


lucab:
Since https://github.com/shepmaster/snafu/pull/12 `snafu` supports capturing backtraces. While useful in several contexts, backtraces comes with additional costs both at runtime (https://github.com/shepmaster/snafu/issues/84) and at buildtime (it uses `backtrace-sys` under the hood, which brings in C source/compilation).

Backtraces are currently active by default and needs to be disabled (opt-out) by consumers via the `default-features = false` syntax.
In the context of an ecosystem of snafu-using libraries, this poses a problem when nesting a couple of levels of dependencies, as the top-level consumer is no more in charge of those features and can't directly opt-out.

Considering that [cargo features are additive](https://github.com/rust-lang/cargo/issues/4328), a better approach would be to make backtraces disabled by default and let the final consumer enable them via a dedicated feature (opt-in).
That way any intermediate library could re-export the `snafu/backtraces` feature, and the top-level application/consumer would be the one (optionally) in charge of enabling backtraces across all crates.

shepmaster:
> as the top-level consumer is no more in charge of those features

But the consumer is not actually in charge of this. The library creating the error already has to add a `Backtrace` type to each corresponding error variant:

```rust
#[derive(Snafu)]
enum Error {
    Something { backtrace: Backtrace }
}
```

With your proposed end result, every crate will also need to 

1. Add a feature flag for backtraces
1. Add conditional compilation to each `backtrace` field:

    ```rust
    #[derive(Snafu)]
    enum Error {
        Something { #[cfg(my-backtraces)] backtrace: Backtrace }
    }
    ```

These multiple steps make me worried that library authors simply won't use backtraces at all, negating any benefits.

There's an inherent tension around default features. Without a feature by default, people get confused because "the docs show it's there, but it doesn't work for me". With a feature by default, time and space can be wasted on unused functionality.

Additionally, there's tension between library usage and application usage. Libraries should strive to be minimal in what they bring in to better work in more contexts, but applications tend to want to be maximal.

Anecdotally, library crates do a poor job of using `default-features = false`.

> and can't directly opt-out.

Can you provide some concrete cases where such a thing would be beneficial? Presumably you are encountering one "in the wild" to prompt the request.

shepmaster:
I wonder if we could do something like

```rust
#[cfg(backtrace)]
struct Backtrace(...);

#[cfg(not(backtrace))]
struct Backtrace;
[...]

PR #121, opened 2019-07-22
Add optional conversion to `backtrace::Backtrace`

### LINKED ISSUE #118: Snafu cannot be used with color_backtrace::print_backtrace
[color-backtrace](https://crates.io/crates/color-backtrace) is a crate for printing colored, expanded backtraces with source context. It offers, among others, a function [print_backtrace](https://docs.rs/color-backtrace/0.2.2/color_backtrace/fn.print_backtrace.html), which takes backtrace::Backtrace as an argument. I wanted to use it with Snafu, however this crate's [backtrace shim](https://docs.rs/snafu/0.4.1/src/snafu/backtrace_shim.rs.html) hides the required type as private.

Would it be possible to somehow expose the inner type to allow usage with color-backtrace?

ArekPiekarz:
This fixes #118

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,9 @@ default = ["rust_1_30", "backtraces"]
 # Adds the backtrace type
 backtraces = ["snafu-derive/backtraces", "backtrace"]
 
+# Add conversion to `backtrace::Backtrace`
+backtrace-crate = ["backtraces"]
+
 # Add extension traits for the futures 0.1 crate
 "futures-01" = ["futures01"]
 
--- a/src/backtrace_shim.rs
+++ b/src/backtrace_shim.rs
@@ -73,6 +73,13 @@ impl fmt::Display for Backtrace {
     }
 }
 
+#[cfg(feature = "backtrace-crate")]
+impl AsRef<backtrace::Backtrace> for Backtrace {
+    fn as_ref(&self) -> &backtrace::Backtrace {
+        &self.0
+    }
+}
+
 struct SymbolDisplay<'a>(&'a backtrace::BacktraceSymbol);
 
 impl<'a> SymbolDisplay<'a> {
--- a/src/guide/feature_flags.md
+++ b/src/guide/feature_flags.md
@@ -11,9 +11,24 @@ cases.
 
 **default**: enabled
 
-When enabled, you can use the [`Backtrace`](crate::Backtrace) type in
+When enabled, you can use the [`Backtrace`] type in
 your enum variant. If you never use backtraces, you can omit this
 feature to speed up compilation a small amount.
+
+[`Backtrace`]: crate::Backtrace
+
+### `backtrace-crate`
+
+**default**: disabled
+
+When enabled, you can convert the SNAFU [`Backtrace`] type to the
+underlying [`backtrace::Backtrace`] type. This allows interoperability
+with other crates that require this type.
+
+It is recommended that only applications make use of this feature. At
+some point in the future, the standard library will have its own
+backtrace type that SNAFU will use and this feature may conflict with
+its use.
 
 ### `unstable-futures`
 


shepmaster:
One thing I realized is that this might conflict with some future plans to make our `Backtrace` type always available but only "active" when the backtrace feature is enabled (see #113 for some background). Conceptually, it would look like:

```rust
#[cfg(not(feature = "backtraces"))]
struct Backtrace;

#[cfg(feature = "backtraces")]
struct Backtrace(backtrace::Backtrace)
```

The same root problem might occur whenever the standard library gains a backtrace type: there wouldn't be a backtrace crate in play then.

I wonder if it's worth building in some fallibility here. Instead of `AsRef`, what if we used some inherent method that could return an `Option`? It kinda sucks that we wouldn't be able to use a nice trait, but I'm not immediately seeing an alternative.

This commit would be more like:

```rust
impl Backtrace {
    #[cfg(feature = "backtrace-crate")]
    fn as_backtrace(&self) -> Option<&backtrace::Backtrace> { Some(/* ... */) }

    // Potentially also `into_backtrace` in the future if it's useful.
}
```

In the future, we would change it to be `None` when using the standard library backtrace or if the user has disabled backtraces.

What do you think?

ArekPiekarz:
If I understood correctly the future direction, you would like the `as_backtrace()` method to be always available, but return `Some` only when the feature `"backtrace-crate"` is enabled. It sounds reasonable to move the burden of conditional compilation logic away from users' code.

shepmaster:
> you would like the `as_backtrace()` method to be always available

Originally, I was going to disagree with this, but it made me think a bit harder on the problem, so thank you!

Maybe the final end state could be:

- The `backtraces` feature is disabled by default
- SNAFU unconditionally has a `Backtrace` type, encouraging error type authors to include it in their types.
- When the application author enables the `backtraces` feature, the SNAFU `Backtrace` type starts to carry a `backtrace::Backtrace`.
- The `backtrace-crate` feature implies `backtraces` and adds the `AsRef` implementation as shown in this PR. It wouldn't need to return an `Option` because of the requirement on `backtraces`.
- In the future, we could add a `backtrace-std` feature that also implies `backtraces`. It would be incompatible with the `backtrace-crate` feature, but only an application should enable it so this shouldn't be a problem. At that point in time, we could also implement `AsRef` for the standard library `Backtrace` type.


For a library author, they just scatter `Backtrace` in all of their leaf errors (and `backtrace(delegate)` in aggregating errors, for now).

An application author adds `features = [backtraces]` and *optionally* one of `backtrace-crate`, `backtrace-std` (sometime), if they need to access the underlying type.

ArekPiekarz:
I've been trying to think of an example where library would need the underlying backtrace type on its own and couldn't find any sensible.

In case of recoverable errors it would return Result and not fiddle with backtraces by itself. In case of unrecoverable ones, like wrong indexing in a Vec, it should use "panic!" and leave optional installing of custom panic handler to the application.

One not really sensible scenario would be a library internally handling an error from a Result and deciding to print the backtrace with other library (like color-backtrace), but that would be a bad design by taking the control away from the application and including unnecessary dependencies.

A reasonable case could be a "snafu-helper" library that takes the Snafu backtrace and somehow works on it and the inner type, but that should be invoked by the application.

All in all I think your design makes sense.

lucab:
@shepmaster the above looks sane to me, except one point that I am not fully following, here below.

> An application author adds features = [backtraces] and optionally one of backtrace-crate, backtrace-std (sometime)

If an application has `features = [backtraces]` only, **without** any further `backtrace-crate` or `backtrace-std`, what does `as_backtrace(&self) -> Option<X>` in terms of the inner type "X"?

shepmaster:
> what does `as_backtrace(&self) -> Option<X>` in terms of the inner type "X"

That function would only be present when `backtrace-crate` or `backtrace-std` is also enabled; those are optional and conflict with each other.

shepmaster:
Thanks!

PR #129, opened 2019-07-27
Make `source(from)` imply `source(true)`

### LINKED ISSUE #124: `source(from)` could imply `source(true)`
In #109, when considering the field attributes that can work together, there was some agreement that `source(true)` and `source(from)` make sense together on a field.  `source(true)` indicates that a field is the source even though it can't be named "source".  `source(from(...))` tells Snafu how to transform another type into the type indicated in the field.  (See https://docs.rs/snafu/0.4.3/snafu/guide/attributes/index.html#transforming-the-source).  There's no conflict in these intentions.

Moreover, if you specify `source(from)` on a field, you almost certainly mean for that to imply `source(true)` - you've already said "source" once, and it's a conflict to specify `source(false)` and `source(from)`.  It's not the case today, though - the `source(from)` attribute would be ignored if there weren't some other signal that the field is a source.

It'd reduce line count and cognitive overhead to have `source(from)` imply that the field is a source.  `source(true)` on the same field would still be valid, just redundant.

tjkirch:
Fixes #124 - based on discussion from #109 about which attributes should be valid together.

This change makes `source(from(...))` imply `source(true)`.  This means you don't need two `source`-type attributes on a field that you intend to be a source but can't be named `source`.  (It's still valid to specify `source(from(...))` and `source(true)`, just redundant.)

The `attributes` section of the user's guide is updated with a note on this behavior.

---

**Testing done:**

Updated compile-fail results show that we catch more duplicates than we could before, because more fields are recognized as sources.

An updated unit test also shows this behavior in a more positive light.  :)

### DIFF
--- a/compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
+++ b/compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
@@ -17,6 +17,12 @@ error: Multiple 'backtrace' attributes are not supported on one field; found a d
    |             ^^^^^^^^^^
 
 error: Multiple 'source' attributes are not supported within an error variant; found a duplicate here:
+  --> $DIR/attribute-duplication.rs:18:13
+   |
+18 |             source2: Box<EnumError>,
+   |             ^^^^^^^
+
+error: Multiple 'source' attributes are not supported within an error variant; found a duplicate here:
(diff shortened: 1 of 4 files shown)


shepmaster:
Looks wonderful!

PR #130, opened 2019-07-27
Add contributor's guide

### LINKED ISSUE #127: Add a contributor's guide
I think it'd be a good idea to have a `CONTRIBUTING.md` in the repo to help guide new contributors.  It can make a project seem a bit more approachable, and can save time in pull requests.

Background on contributor's guides on GitHub: https://help.github.com/en/articles/setting-guidelines-for-repository-contributors

I picked up a few specific tips from #109 that could serve as starting material; something like these could be combined with more general procedural tips and some encouragement for newcomers.

Communication tips:
* Open an issue for discussion before writing any non-trivial changes - often the author or other contributors can help shape an even better idea.  Or maybe someone is already working on it!
* Even if you're making a breaking change, don't worry about updating the version number or changelog -- they're done together before a release -- but feel free to suggest wording you like in the pull request.
* We value correctness and clarity in the code, API, and docs, and it's worth putting in the time for thorough review in issues and pull requests.
* Don't try to fix the world in a single issue or pull request.  Even small issues can sprout many good ideas, and feel free to split those into new issues.

Testing tips:
* We maintain compatibility with older versions of Rust, and this is enforced through compatibility testing that runs automatically when you create or update a pull request.  You can run these earlier, locally, by running `cargo test` in one of the directories under `compatibility-tests/` - the `rust-toolchain` files there will cause the right version of Rust to be used, assuming you use rustup.  Be aware that you may need to install specific versions of dependencies, as described in `.cirrus.yml` at the root of the repo.
* If you're adding a new compile-time error, add a sample under `compatibility-tests/compile-fail/tests/ui/` to be sure it fails in the way you expect.
* If you're adding a new feature, add a unit test for it in a file that sounds relevant under `tests/` - unit tests are not written in the library files.

General tips:
* Before 1.0, breaking changes are OK if they're adding value - we can update the minor version as needed.
* The user's guide is a valuable resource and worth the time to keep up to date when adding or changing the library.

tjkirch:
Fixes #127 

This is some initial wording that I hope can help new contributors to SNAFU.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,43 @@
+# Contributor's Guide
+
+Thank you for helping with SNAFU!
+
+This document has some guidelines and tips that can help you make a contribution.
+Feel free to make a pull request to this file, too, if you learn anything during your contribution that can help others.
+
+## Code of Conduct
+
+This project is governed by the [Code of Conduct](https://github.com/shepmaster/snafu/blob/master/CODE_OF_CONDUCT.md).
+Please understand those guidelines, and report violations to @shepmaster.
+
+## Getting Started
+
+If you're looking for a way to contribute - first of all, thanks!
+Here are some ideas:
+
+* Issues that we need help on are tagged [help wanted](https://github.com/shepmaster/snafu/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)
+* Issues good for beginners are tagged [good first issue](https://github.com/shepmaster/snafu/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)
+* Anything that's not clear to you in the documentation, particularly in the [user's guide](https://docs.rs/snafu/latest/snafu/guide/index.html)
+
+## Communication tips
+
+* Open an issue for discussion before writing any non-trivial changes.  Often the author or other contributors can help shape an even better idea.  Or maybe someone is already working on it!
+* Even if you're making a breaking change, don't worry about updating the version number or changelog.  They're done together before a release.  Feel free to suggest some wording you like in the pull request, though.
+* We value correctness and clarity in the code, API, and docs, and it's worth putting in the time for thorough review in issues and pull requests.
+* Don't try to fix the world in a single issue or pull request.  Even small issues can sprout many good ideas, and feel free to split those into new issues.
+
+## Testing tips
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+* We value correctness and clarity in the code, API, and docs, and it's worth putting in the time for thorough review in issues and pull requests.
+* Don't try to fix the world in a single issue or pull request.  Even small issues can sprout many good ideas, and feel free to split those into new issues.
+
+## Testing tips
+
+* We maintain compatibility with older versions of Rust, and this is enforced through compatibility testing that runs automatically when you create or update a pull request.  You can run these earlier, locally, by running `cargo test` in one of the directories under `compatibility-tests/`.  The `rust-toolchain` files there will cause the right version of Rust to be used, assuming you use rustup.  The `Cargo.toml` files there will make sure that compatible dependency versions are used, too.
+* If you're adding a new compile-time error, add a sample under `compatibility-tests/compile-fail/tests/ui/` to be sure it fails in the way you expect.
+* If you're adding a new feature, add a unit test for it in a file that sounds relevant under `tests/`.  (Unit tests are not written in the library files in SNAFU.)
```
shepmaster: > Unit tests are not written in the library files in SNAFU

I don't have any objection to unit tests *per se*; in fact I'd prefer them where possible. The problem is that it's very hard to have any **unit** tests because most interesting things to test are in `snafu-derive`. We could have unit tests there with raw strings containing Rust source code, but that felt like overkill.

Instead, I've mostly focused on integration tests, which are the types of tests present in `tests/`.

That's my reasoning; do you feel like any of that is useful to include here?

CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+
+## General tips
+
+* Before 1.0, breaking changes are OK if they're adding value - we can update the minor version as needed.
```
shepmaster: I'm actually OK with breaking changes in general. I don't want to add them for no reason, but I don't have an aversion to releasing 2.x, 3.x, etc. 

I think the main difference is the time scale — 0.x to 0.(x+1) could be a month or so, but x.0 to (x+1).0 should be 9+ months. (In general, when there aren't really good reasons, prices and participation may vary).

Is any of that is useful to include here?

CONTRIBUTING.md
tjkirch: Ah, ok, I was misunderstanding the purpose.  I was thinking of some of the tests under `tests/` as unit tests, and made the leap that unit tests were put there instead of in the library files directly.  I'll try to clarify this bullet point.

CONTRIBUTING.md
tjkirch: > Is any of that is useful to include here?

Definitely!  I'll try to clarify, thanks.

tjkirch:
I addressed the comments about unit testing and breaking changes - please let me know what you think.  :)

shepmaster:
Feel free to merge when CI passes ;-)

PR #133, opened 2019-07-31
Improve accuracy of error spans

### LINKED ISSUE #126: Improve accuracy of spans in attribute misuse/duplication errors
In #109, new error messages were added to help explain misuse and duplication of snafu attributes.  The source spans used for the error messages refer to the name of the struct, enum, variant, or field on which the attribute was placed.  This is helpful for finding the issue, but isn't the most accurate - in some cases it would be better to point to the attribute itself, rather than the element it was placed on.

tjkirch:
```
    Improve accuracy of error spans for attribute duplication and misuse
    
    `syn::Error::new_spanned` can pinpoint erroneous attributes using `syn` token
    trees, which we now store in `SnafuAttribute`. This lets us give them to
    `SyntaxErrors` and `AtMostOne` for later reporting.
```

---

Fixes #126

Change details:
* Use `syn::Error::new_spanned` with the token tree of the attribute, instead of `syn::Error::new` with the span; this results in errors pointing to the relevant attribute instead of the field or an odd subset of the attribute text.
* Simplify error messages because they no longer have to work harder to explain the awkward spans.
* In `AtMostOne`, store the location (token tree) with the value, which allows for a new `finish_with_location` method that returns the value and location instead of just the value.  This allows using the original location for accurate error messages resulting from additional manual checks.
* Modify `SnafuAttribute` to store the location as well as the internal data.

What I like:
* The errors are more direct, and no longer dedupe because of multiple errors on one element.
* I think `finish_with_location` is a good mirror to `finish`, and it's worth having both because it expresses intent and (if you only had `finish_with_location`) there are cases where it's awkward to ignore the extra field.
* No longer need the "is_*" `Option<bool>` fields to track field type; it's (arguably) more naturally handled with the updated `AtMostOne`, though we do still need bools to track the `(false)` attributes.

What I don't like:
* I had to clone some of the `TokenStream`s when they were being used in multiple errors or `AtMostOne`s.
  * I couldn't use references because we're generally adding elements/errors that we want to use after looping through the attrs/locations from `attributes_from_syn`, and they'd drop at the end of the loop.
  * I couldn't figure out a `Box` strategy because I think it'd involve the same cloning; with the current general structure, we need to store the same `TokenStream` multiple times, like in the inner loop over Source elements (`ss`) that use the same `tts`, or when we store the location in `source_attrs` to check for high-level duplicates of "source" *and* in `transformations` to track the more-specific `source(from)` details.  I think we'd just need two boxes with two clones of the data...
* Some minor things were made uglier by the need to store the extra data, rather than being able to spontaneously generate a span.
  * The `(false)` tracking mentioned above.
  * Rust 1.18 wasn't happy about destructuring `Option`s carrying tuples, which resulted in a couple `.1` references instead of the destructuring that works in the 1.30+ compat tests.

Overall, I think it resulted in nice clarity improvements for not too much extra complexity, but I'd be happy for feedback on improving the points above!

---

**Testing done:**

Existing integration and compatibility tests pass.

The compile-fail tests were updated with the more accurate errors; they were reordered quite a bit, but I'm fairly sure it's catching the same cases, just in different ways.  There's also a backtrace-related error that deduped away before, but now that the errors point to the duplicate attributes themselves, on different lines, it shows up.


snafu-derive/src/lib.rs
```diff
@@ -203,17 +204,24 @@ impl fmt::Display for DuplicateAttribute {
 /// it's added to a list of DuplicateAttribute errors, using the given `name` and `location` as
 /// descriptors.
 ///
-/// When done parsing a structure, call `finish`; it gives you the first attribute found, if any,
-/// and the list of errors.
+/// When done parsing a structure, call `finish` to get first attribute found, if any, and the list
+/// of errors, or call `finish_with_location` to get the attribute and the token tree where it was
```
shepmaster: I'm surprised that there are times where we want to call just `finish` and *not* care about the error location...

snafu-derive/src/lib.rs
```diff
@@ -451,31 +471,34 @@ fn parse_snafu_enum(
...
-                // down to one error; it should be enough to find the problem.
                 let mut source_attrs = AtMostOne::new("source", "on one field");
+                let mut transformations = AtMostOne::new("source(from)", "on one field");
                 let mut backtrace_attrs = AtMostOne::new("backtrace", "on one field");
 
-                for attr in attributes_from_syn(syn_field.attrs)? {
+                // Keep track of the negative markers so we can check for inconsistencies and
+                // exclude fields even if they have the "source" or "backtrace" name.
+                let mut has_source_false = false;
+                let mut has_backtrace_false = false;
```
shepmaster: Ooof. Double negatives make my head hurt. I don't have the brainpower right now to think of an alternative, but I'd like to see if there's something we can come up with.

snafu-derive/src/lib.rs
```diff
@@ -985,18 +1020,35 @@ impl syn::parse::Parse for DocComment {
...
 }
 
-fn attributes_from_syn(attrs: Vec<syn::Attribute>) -> MultiSynResult<Vec<SnafuAttribute>> {
+/// Given a list of syn `Attribute`, parses each attribute into a type known to SNAFU, namely a
+/// variant of SnafuAttribute.  Returns a list of pairs, where each pair is a SnafuAttribute and
+/// the TokenStream where it was found; the TokenStream can be used to give an accurate error
+/// message if there's a problem with the attribute.
+fn attributes_from_syn(
+    attrs: &[syn::Attribute],
+) -> MultiSynResult<Vec<(SnafuAttribute, proc_macro2::TokenStream)>> {
```
shepmaster: What do you think about putting the `TokenStream` inside the `SnafuAttribute` itself, instead of returning the tuple?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
```diff
@@ -1,95 +1,107 @@
...
+   |                    ^^^^^^^^^^^
 
-error: Multiple 'source' attributes are not supported within an error variant; found a duplicate here:
-  --> $DIR/attribute-duplication.rs:25:13
+error: Multiple 'source' attributes are not supported within an error variant
+  --> $DIR/attribute-duplication.rs:16:20
    |
-25 |             source4: String,
-   |             ^^^^^^^
+16 |             #[snafu(source(from(EnumError, Box::new)))]
```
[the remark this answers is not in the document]
tjkirch: The order comes the fact that I'm extending the overall error list with the errors from each `SyntaxErrors`/`AtMostOne` as I `finish` it.  I think the way to achieve overall ordering would be to store the line number with each element as we add it, and then sort the final error list by that number.  The line number can be retrieved through `Spanned` on the attribute, then `.start` to get `LineColumn`.`line`.  Do you think that's worthwhile?

snafu-derive/src/lib.rs
tjkirch: I thought so too, and originally I just had `finish_with_location`, but I found that throwing away the extra field in the tuple was unergonomic in some cases, and it seemed like minimal overhead to have a function for each case.  There are 6 uses of `finish` which felt sufficient to me.  But if it stood out to you too, maybe I should try again?

snafu-derive/src/lib.rs
tjkirch: The original `is_source`, `is_backtrace`, and `is_backtrace_delegate` were certainly clear in their intention, but you had to be (IMHO) more careful about when to flip and check those.  Using the existing `AtMostOne`s felt natural since they were already tracking which attributes we'd seen, and it eliminated duplicated state.  That just left the "odd" case of a user denying that the field is what it appears to be.  An alternative could be to add a method to `AtMostOne` that lets you override what it's seen, guaranteeing it will only give errors upon `finish`.  That doesn't feel particularly clean to me either, though.  Hmm.

snafu-derive/src/lib.rs
tjkirch: I had considered that, and it seemed like the best approach would be to replace `SnafuAttribute` with a struct that carries the location and the enum currently called `SnafuAttribute`, rather than putting the location in each variant of the current enum.  In the end I went with the modification to `attributes_from_syn` because the returned tuples felt natural to me, and it didn't require modifying every use of `SnafuAttribute` to add a field indirection.  I don't feel strongly, though.  What do you think is best of those?  Or do you have other implementation ideas that would simplify storing location with the `SnafuAttribute`?

tjkirch:
> the highlighted span is "one bigger" than I would have expected

Same here, and I haven't been able to figure it out yet, though I haven't dived in fully.  I *think* this `impl ToTokens for Attribute` is what would be generating it, but I'm having trouble understanding it.

https://docs.rs/syn/0.15.42/src/syn/attr.rs.html#651

Maybe it's something to do with `syn::token::Bracket` and the `surround` method called in that code?

> I wonder what would be highlighted for `#[snafu(visibility(pub), visibility(pub))]`

This doesn't seem to work for me; is this expected to work?  I tried a couple different ways with different combinations of attributes.  (I admit I've been confused when dealing with the code that can handle multiple attributes inside a `#[snafu()]`.)

```
error: unexpected token
 --> $DIR/attribute-duplication.rs:6:28
  |
6 |     #[snafu(visibility(pub), visibility(pub))]
  |
```

```
error: unexpected token
 --> $DIR/test.rs:6:44
  |
6 |         #[snafu(display("an error variant"), visibility(pub))]
  |
```

(It got snipped at bit, but the unexpected tokens are the commas)

snafu-derive/src/lib.rs
tjkirch: I thought so too, and originally I just had `finish_with_location`, but I found that throwing away the extra field in the tuple was unergonomic in some cases, and it seemed like minimal overhead to have a function for each case. There are 6 uses of `finish` which felt like a good case to me in the end. But if this stood out to you too, maybe I should try again?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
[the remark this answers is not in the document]
shepmaster: I don't know that it's *super* worth it, mostly because the error test is a pathological case — I certainly hope that people aren't adding that many wrong / duplicate attributes at once! 

The main benefit I see is that our test output would be more stable. Maybe just a small thing to do in the future, if it becomes annoying.

shepmaster:
> is this expected to work

Hmm, yes, it's *supposed to*, and I also see that it doesn't. That's the intent of https://github.com/shepmaster/snafu/blob/31ac98aa3b4aff990fc8c5f416489f485cebf8cb/snafu-derive/src/lib.rs#L960-L972

snafu-derive/src/lib.rs
shepmaster: > replace `SnafuAttribute` with a struct that carries the location and the enum currently called `SnafuAttribute`

Yeah, the pattern of "an enum where every variant has the same piece of data" is a rather annoying case. I flip between the solution you mentioned and

```rust
enum E {
    A(Same, Diff1),
    B(Same, Diff2,
    C(Same, Diff3),
}

impl E {
    fn same(&self) -> &Same { /* ... */ }
}
```

I'm OK with the tuples for now, but I have the feeling that if it grows that the pieces will want names, growing into the enum-wrapped-struct you mentioned.

snafu-derive/src/lib.rs
shepmaster: > 6 uses of `finish` which felt like a justification to me in the end

Yes, so long as there are cases where the information isn't needed / used, it makes sense. My question is mostly around *why* isn't it needed in those cases, why was it being thrown away?

tjkirch: > My question is mostly around why isn't it needed in those cases, why was it being thrown away?

Most error handling is done inside `AtMostOne` now, and the errors are surfaced when `finish` is called so they can be added to a `SyntaxErrors` or returned directly.  The location is useful for manual error checks, when we want to check for a condition other than there being more than one of an attribute, for example checking whether two incompatible attributes are used.  In those cases, you need the location to manually add an error.  In most cases, though, the location handling is done when `add`ing things to an `AtMostOne`.

An alternative could be to abstract `AtMostOne` into a more general error-handling structure that can check for duplication and other types of errors like incompatibility, but to me it seems like that would be quite a bit of complexity to handle the relatively few non-duplication errors we have now.  (I'm not counting attribute misuse errors, because in those cases you can immediately add an error to the list and don't need any higher-level tracking.)

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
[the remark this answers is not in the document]
tjkirch: > our test output would be more stable

I'm curious what you mean on this one - I'd say the errors are "stable" in that they're always generated in the same order, and so can be consistently tested.  They're not ordered by line number, something that would be helpful for humans looking at a full error listing.  (And I agree, hopefully people aren't adding so many errors that this would be a problem...)

tjkirch:
> I wonder what would be highlighted for `#[snafu(visibility(pub), visibility(pub))]`

With #137 fixing the parsing of multiple attributes, I can confirm that both of the errors generated for a line like this point to the entire parenthesized expression rather than the relevant inner attribute.  :(

```
error: 'visibility' attribute is only valid on an error enum and its variants, not on a field
  --> src/main.rs:15:20
   |
15 |             #[snafu(visibility(pub), display("Moo"))]
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: 'display' attribute is only valid on variants of an error enum, not on a field
  --> src/main.rs:15:20
   |
15 |             #[snafu(visibility(pub), display("Moo"))]
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I was hoping it'd be something obvious like a `parenthesized!` in the wrong place, but it doesn't seem quite that easy.  Maybe a missing `parenthesized!`...  I'm trying to figure this out now.

tjkirch:
Ah - I think the issue where error pointers include the outer parentheses is caused by `attributes_from_syn` taking the `TokenStream` from the `syn::Attribute` instead of the parsed `SnafuAttribute`.  I traced through the syn parsing code and everything seemed to be correct, and then realized that `attributes_from_syn` is ignoring all the careful parentheses-handling in the `impl syn::parse::Parse` code by storing the `TokenStream` from before that parsing is done.

So, I think shepmaster's comment about "putting the `TokenStream` inside the `SnafuAttribute`" is doubly relevant; if we can store the `TokenStream` inside, it should point to just that attribute, and then we can use that more accurate attribute where needed.  And without returning the location separately from `attributes_from_syn`.

tjkirch:
> if we can store the TokenStream inside, it should point to just that attribute, and then we can use that more accurate attribute where needed

I'm making progress on this.  I've got some of the errors pointing inside the parentheses.  It's not quite right yet, though, because some are pointing *too far* inside, for example just the "pub" in "snafu(visibility(pub))" on a visibility misuse.  Getting close though.

I tried SnafuAttribute as a struct and enum, and it was cleaner as an enum with a common field and a method to pull out that common field.  It also required that I add the common field (a TokenStream) to MyMeta and MyParens, because we need access to the location *after* they deal with parentheses.

shepmaster:
It's unfortunate that we can't "take a snapshot" when we start in `impl syn::parse::Parse for SnafuAttribute` and then take another when we finish, reporting that as the span.

tjkirch:
> It's unfortunate that we can't "take a snapshot" when we start in `impl syn::parse::Parse for SnafuAttribute` and then take another when we finish, reporting that as the span.

Hehe... actually, I was, but I was only using the "snapshot" span for some cases.  Turns out I was just overcomplicating the other cases, and the input span is exactly what I wanted.  Thanks!  I should be able to update the PR today.

tjkirch:
I pushed the fix for the "one bigger" spans, and marked this PR ready for review.  There were still a couple design/style comments open above, which I tried to answer as best I can, but I think overall it's in good shape for consideration.

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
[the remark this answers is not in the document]
shepmaster: I mostly mean that the order of things in this standard error file are based on the order of the code. If we do some "trivial" refactoring and flip the order of two calls, then this file will be invalidated, even though the errors are the same.

I also know that as this file grows, we will get a lot of incidental diffs; line numbers changing because we added/removed a line at the top. If that proves to be a problem, I might split it out into multiple files just to reduce that kind of churn.

compatibility-tests/compile-fail/tests/ui/attribute-misuse.stderr
```diff
@@ -1,77 +1,77 @@
...
+  --> $DIR/attribute-misuse.rs:15:21
    |
-18 |             a: String,
-   |             ^
+15 |             #[snafu(visibility(pub))]
+   |                     ^^^^^^^^^^^^^^^
 
-error: Incompatible attributes ["source(false)", "source(from)"] specified on one field here:
-  --> $DIR/attribute-misuse.rs:18:13
+error: Incompatible attributes ["source(false)", "source(from)"] specified on one field
```
shepmaster: Could we update it to say "on a field"?

snafu-derive/src/lib.rs
```diff
@@ -259,6 +267,17 @@ impl<T> AtMostOne<T> {
     /// Consumes the AtMostOne, returning the first item added, if any, and the list of errors
     /// representing any items added beyond the first.
     fn finish(self) -> (Option<T>, Vec<syn::Error>) {
+        let errors = match self.errors.finish() {
+            Ok(()) => Vec::new(),
+            Err(vec) => vec,
+        };
+        (self.value.map(|(val, _tts)| val), errors)
```
shepmaster: What do you think about calling `finish_with_location` here?

```rust
let (name1, name2) = self.finish_with_location();
(name1.map(|(val, _tts)| val), name2)
```

snafu-derive/src/lib.rs
```diff
@@ -903,45 +938,53 @@ impl syn::parse::Parse for Backtrace {
...
+    Source(proc_macro2::TokenStream, Vec<Source>),
+    Backtrace(proc_macro2::TokenStream, Backtrace),
+    DocComment(proc_macro2::TokenStream, String),
 }
 
 impl syn::parse::Parse for SnafuAttribute {
     fn parse(input: syn::parse::ParseStream) -> SynResult<Self> {
         use syn::{Expr, Ident, Visibility};
 
+        let input_tts = input.cursor().token_stream();
```
shepmaster: Huh. I saw these functions in syn, but didn't realize that they could be used this way. Cool!

snafu-derive/src/lib.rs
shepmaster: What about using something like "opt out"? Maybe like `source_opt_out = false`?

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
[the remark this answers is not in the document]
tjkirch: Ahh, I understand.  The big compile-fail tests for this bugged me anyway, so I'll split those out, and hopefully that will address the ordering fairly well too.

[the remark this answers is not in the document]
shepmaster: To clarify, I mean that the order of the code *in snafu-derive* will affect the order of the generated errors.

You don't have to split out the compile-fail tests into multiple right now if it's not relevant; I mostly gave that as another example of the kind of "instability" of these tests

---

Truth be told, I'm not a big fan of this style of test, where the entire snapshot is captured. I find them overly brittle in general, but I wanted to use the better-supported trybuild library, which only has snapshots.

tjkirch:
> split it out into multiple files just to reduce that kind of churn

I split out the duplication cases into small sample files that make it easier to understand the different types of errors.  I left the misuse cases in a single file, since they're more straightforward, and they're unlikely to change order because they're added to the top-level error list immediately when they're found. (I did clean up the misuse file a bit, though.)

I also added the other suggestions:
* backticks instead of quotes on attribute names in errors  (via the Display implementation for simplicity)
* "a field" instead of "one field"
* reusing `finish_with_location` in `finish`
* "opt_out" variable names

compatibility-tests/compile-fail/tests/ui/attribute-duplication.stderr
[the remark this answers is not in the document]
tjkirch: > To clarify, I mean that the order of the code in snafu-derive will affect the order of the generated errors.

Yep, I realized that part after your hint in the last message.  I split out the tests so that there aren't multiple unrelated error checks being tested in one file, which means that reordering the code in snafu-derive won't unnecessarily break one of the compile-fail tests.  I also think it made it easier to understand which error case was being checked, when there aren't a bunch of other attributes/variants in the same test case.  (Test cases can also be split into modules for clarity, but that doesn't help with the irrelevant changes to line numbers in the snapshots.)

compatibility-tests/compile-fail/tests/ui/duplication-enum.rs
```diff
@@ -0,0 +1,11 @@
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub))]
+#[snafu(visibility(pub))]
+enum EnumError {
+    #[snafu(display("an error variant"))]
```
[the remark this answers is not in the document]
tjkirch: Honestly, these were all because I didn't know `display` was optional :)

compatibility-tests/compile-fail/tests/ui/duplication-source-from-attributes.stderr
```diff
@@ -0,0 +1,11 @@
+error: Multiple `source` attributes are not supported on a field
+ --> $DIR/duplication-source-from-attributes.rs:9:18
+  |
+9 |          #[snafu(source(from(EnumError, Box::new)))]
+  |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: Multiple `source(from)` attributes are not supported on a field
+ --> $DIR/duplication-source-from-attributes.rs:9:18
+  |
+9 |          #[snafu(source(from(EnumError, Box::new)))]
```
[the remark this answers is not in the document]
tjkirch: It's because I used an `AtMostOne` called `source_attrs` for `source` tracking, and another called `transformations` for `source(from)` tracking.  When hitting a `source(from)` I need to add to both:

https://github.com/shepmaster/snafu/blob/88ee0375ac9ec84beb2fd25982e4d10ef057dd37/snafu-derive/src/lib.rs#L536-L537

This is so we can understand that a `source(from)` also counts as a source field, rather than allowing two source fields just because one uses `source` (or is named "source") and one uses `source(from)`.

An alternative would be to use a single list of source attributes, and track whether it was a transformation (`source(from)`) inside of each list element.  Currently `source_attrs` is just storing `()` as its data element, so maybe that could be an `Option` of the transformation data, for example.  I can try that.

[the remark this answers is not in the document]
shepmaster: Oh! I missed that one was `source` and one was `source(from)`; that's fine, carry on!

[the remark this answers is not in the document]
tjkirch: On balance, I think the "duplicate" is confusing and you were right to point it out.  It's caused by the same line of the user's code, and looks almost identical, and was only there because of an implementation detail.  I don't think it added any clarity to the presented errors.

Plus, the code is simpler without it, modulo some Rust 1.18 complaints.  I'm going to propose it in the next push and see what you think.  :)

tjkirch:
This push:
* Moves the test for duplicate attributes occurring on one line to a new file
* Removes irrelevant `display` attributes in tests
* Stores transformation info inside `source_attrs`, rather than a separate `AtMostOne`
  * ...which removes a duplicate error in duplication-source-from-attributes.stderr
  * ...and which required we store all values seen by the `AtMostOne` instead of just the latest
    * This uses more memory, but simplified the code, and allowed for useful things like the above
* Leaves the inner type implied in `impl fmt::Display for IncompatibleAttributes`
* Uses backticks instead of quotes in the error regarding using `backtrace` and `backtrace(delegate)` fields in the same error variant

tjkirch:
> ...and which required we store all values seen by the AtMostOne instead of just the latest

This was too strong a statement on my part, since we could do the necessary tracking in this case with a boolean, but I'm trying to move away from booleans :)  Exposing an iterator of values seemed like an improvement to both clarity and generality, and opens up other possibilities for error checking, so I liked that.

shepmaster:
> I'm trying to move away from booleans

Yeah, booleans are like most other primitives: great when they are in a pretty narrow context, terrible once they are spread out. Giving fancy names to things is 👌 

> This uses more memory

I fall into this line of thinking as well, but it's helpful to remember that we are unlikely to ever storing even 10 items, so it's not like a huge waste. (I wonder if there's a way to get the memory usage of a proc-macro...)

snafu-derive/src/lib.rs
```diff
@@ -203,67 +210,91 @@ impl fmt::Display for DuplicateAttribute {
...
+    fn finish_with_location(mut self) -> (Option<(T, U)>, Vec<syn::Error>) {
         let errors = match self.errors.finish() {
             Ok(()) => Vec::new(),
             Err(vec) => vec,
         };
-        (self.value, errors)
+        // We're done with the `AtMostOne` and want to return the first element, if any.
+        // Truncate to throw away any extras, and pop to get an Option with (maybe) the first.
+        self.values.truncate(1);
+        (self.values.pop(), errors)
```
shepmaster: In the realm of potentially micro-optimizing, you could use a `VecDeque` here, but I wouldn't worry about it ;-)

tjkirch: Ah, regardless of optimization, it's clearer for the purpose!  Switching to `VecDeque`.

Issue #134, opened 2019-08-01
snafu::futures::try_future::Context is not awaitable


pluth:
I'm trying to use snafu in an async/await project and I am running into the following Issue:

Every Future (std) implements TryFuture when it returns Result<_, _> due to the blanket implementation in https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.17/futures/future/trait.TryFuture.html#implementors. However, the opposite does not seems to be true, i.e. a TryFuture does not automatically implement Future.

As snafu::futures::try_future::Context implements TryFuture directly, it does not implement Future and is therefore not awaitable. Would you consider a change to Context so that it implements Future<Output = Result<..>> instead? It would keep the TryFuture implementation through the blanket implementation above.

I've modified the example of the [context](https://docs.rs/snafu/0.4.3/snafu/futures/trait.TryFutureExt.html#tymethod.context) method to demonstrate:

```rust
#![feature(async_await)]

use futures::future::TryFuture;
use snafu::{futures::TryFutureExt, Snafu};

#[derive(Debug, Snafu)]
enum Error {
    Authenticating {
        user_name: String,
        user_id: i32,
    },
}

async fn example() -> Result<i32, Error> {
    another_function().context(Authenticating {
        user_name: "admin",
        user_id: 42_i32,
    }).await // Error here
   // the trait bound `snafu::futures::try_future::Context<impl core::future::future::Future, 
   // &Authenticating<&str, i32>, Error>: core::future::future::Future` is not satisfied
   // required by `std::future::poll_with_tls_context`rustc(E0277)
}


async fn another_function() -> Result<i32, ()> {
    unimplemented!();
}

```

shepmaster:
> Would you consider a change to Context so that it implements Future<Output = Result<..>> instead

Certainly; so long as it continues to pass what tests we have, ideally adding a new test for the functionality that is currently lacking.

TBH, I wasn't sure how people would want to call it. I'd have expected that most usages would be `future.await.context(...)` (using `ResultExt`) instead of `future.context(...).await` (using `TryFutureExt`);

Any change would also want to apply to the `Stream` combinators as well.

pluth:
> I'd have expected that most usages would be future.await.context(...) (using ResultExt)

Yes, that's what I'm doing now. I just tried TryFutureExt first, mostly because I thought it would look better with await at the end.

I will create a PR

shepmaster:
> I thought it would look better with await at the end.

[...]

PR #136, opened 2019-08-01
Implement Future for Context instead of TryFuture

pluth:
Implements the changes discussed in #134 
Implements Future for try_future::Context and try_future::WithContext instead of TryFuture. TryFuture implementation should still be available as it was through a blanket implementation in futures_core. Same thing for the try_stream module.

I've extended the futures compatibility test to test for the Future implementation (very uncreatively) and removed two into_stream() calls to demonstrate Stream implementation.

### DIFF
--- a/compatibility-tests/futures/src/lib.rs
+++ b/compatibility-tests/futures/src/lib.rs
@@ -61,6 +61,21 @@ async fn load_stock_data_concurrent() -> Result<String, Error> {
     Ok(format!("{}+{}", apple, google))
 }
 
+// Return values of the combinators implement `Future`
+async fn load_stock_data_sequential_again() -> Result<String, Error> {
+    let apple = api::fetch_page("apple")
+        .context(UnableToLoadAppleStock)
+        .await?;
+
+    let google = api::fetch_page("google")
+        .with_context(|| UnableToLoadGoogleStock {
+            name: String::from("sequential"),
+        })
+        .await?;
+
+    Ok(format!("{}+{}", apple, google))
+}
+
 // Can be used as a `Stream` combinator
 async fn load_stock_data_series() -> Result<String, Error> {
     let apple = api::keep_fetching_page("apple").context(UnableToLoadAppleStock);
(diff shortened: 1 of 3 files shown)


compatibility-tests/futures/src/lib.rs
```diff
@@ -61,14 +61,29 @@ async fn load_stock_data_concurrent() -> Result<String, Error> {
     Ok(format!("{}+{}", apple, google))
 }
 
+// Returns of the combinators implement `Future`
```
shepmaster:
```suggestion
// Return values of the combinators implement `Future`
```

compatibility-tests/futures/src/lib.rs
```diff
@@ -61,14 +61,29 @@ async fn load_stock_data_concurrent() -> Result<String, Error> {
...
+}
+
 // Can be used as a `Stream` combinator
 async fn load_stock_data_series() -> Result<String, Error> {
     let apple = api::keep_fetching_page("apple").context(UnableToLoadAppleStock);
     let google = api::keep_fetching_page("google").with_context(|| UnableToLoadGoogleStock {
         name: String::from("stream"),
     });
 
-    let together = apple.into_stream().zip(google.into_stream());
```
shepmaster: I'm not following why this change was made; was it *required* to change or is it simply clearer to change it?

src/futures/try_stream.rs
```diff
@@ -3,6 +3,7 @@
 //! [`TryStream`]: futures_core::TryStream
 
 use futures_core::stream::TryStream;
+use futures_core::stream::Stream;
```
shepmaster: Let's go ahead and do `use futures_core::stream::{Stream, TryStream};`

compatibility-tests/futures/src/lib.rs
shepmaster: Ah, right. `into_stream` is the hack needed to convert a `TryStream` back into a `Stream`, which we don't need anymore.

However, leaving the method calls *would* still work, since we implicitly implement `TryStream`.

shepmaster:
Thanks!

pluth:
Thanks for fixing and implementing it. For me, it can wait. As we discussed, I can just work with the Result combinators

Thanks again

shepmaster:
> For me, it can wait

Too bad! I just released 0.4.4 with this and another fix 😉

PR #138, opened 2019-08-06
Implement Support for no_std

### LINKED ISSUE #85: How can we better support no-std usecases?
Originally, SNAFU was going to support no-std, until I learned that the `Error` trait doesn't exist in libcore!

That being said, [people have said that failure works will in no-std](https://www.reddit.com/r/rust/comments/boyijw/errors_and_best_practises_should_i_create_custom/enns4vv/). Evidently, failure can just implement `Display` in this case.

We certainly could do the same, but I wonder if there's some other powers we could add. My wildest thought would be to establish a parallel `Error` trait that we could implement instead of (or in addition to?) `std::error::Error`.

Would this be something that `ErrorCompat` should do?

CryZe:
This brings no_std support to snafu. It now has a `std` feature which is activated by default. To make no_std support as frictionless as possible, snafu now internally either uses the `std::error::Error` trait when the feature is activated and defines its own API compatible trait instead when it's disabled, but it is never exposed publicly for now.

Resolves #85

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -27,6 +27,22 @@ stable_test_task:
     - cargo fmt --all -- --check
   before_cache_script: rm -rf $CARGO_HOME/registry/index
 
+no_std_test_task:
+  name: "Rust Stable (no_std)"
+  container:
+    image: rust:latest
+    cpu: 1
+    memory: 2Gi
+  cargo_cache:
+    folder: $CARGO_HOME/registry
+    fingerprint_script: cat Cargo.toml
+  setup_script:
+    - rustup target add thumbv6m-none-eabi
+  primary_test_script:
+    - rustc --version
+    - cargo build --no-default-features --target thumbv6m-none-eabi
+  before_cache_script: rm -rf $CARGO_HOME/registry/index
+
 nightly_test_task:
   name: "Rust Nightly"
   container:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ repository = "https://github.com/shepmaster/snafu"
 
 license = "MIT OR Apache-2.0"
 
-keywords = ["error", "ergonomic", "library"]
-categories = ["rust-patterns"]
+keywords = ["error", "ergonomic", "library", "no_std"]
+categories = ["rust-patterns", "no-std"]
 
 exclude = [
   "/.cirrus.yml",
@@ -24,10 +24,13 @@ exclude = [
 all-features = true
 
 [features]
-default = ["backtraces"]
+default = ["backtraces", "std"]
+
+# Implement the `std::error::Error` trait.
+std = []
 
 # Adds the backtrace type
-backtraces = ["snafu-derive/backtraces", "backtrace"]
+backtraces = ["std", "snafu-derive/backtraces", "backtrace"]
 
 # Add conversion to `backtrace::Backtrace`
 backtrace-crate = ["backtraces"]
--- a/src/no_std_error.rs
+++ b/src/no_std_error.rs
@@ -1,0 +1,45 @@
+#![allow(missing_docs)]
+
+use core::fmt::{Debug, Display};
+
+pub trait Error: Debug + Display {
+    fn description(&self) -> &str {
+        "description() is deprecated; use Display"
+    }
+    fn cause(&self) -> Option<&dyn Error> {
+        self.source()
+    }
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        None
+    }
+}
+
+macro_rules! impl_error {
+    ($($e:path),*) => {
+        $(
+            impl Error for $e {}
+        )*
+    }
+}
+
+// All errors supported by our minimum suported Rust version can be supported by
+// default.
+impl_error![
+    core::str::ParseBoolError,    // 1.0
+    core::str::Utf8Error,         // 1.0
+    core::num::ParseIntError,     // 1.0
+    core::num::ParseFloatError,   // 1.0
+    core::char::DecodeUtf16Error, // 1.9
+    core::fmt::Error,             // 1.11
+    core::cell::BorrowMutError,   // 1.13
+    core::cell::BorrowError,      // 1.13
+    core::char::ParseCharError    // 1.20
+];
+
+// We can gate these together with std futures.
+#[cfg(feature = "unstable-futures")]
+impl_error![
+    core::num::TryFromIntError,     // 1.34
+    core::array::TryFromSliceError, // 1.34
+    core::char::CharTryFromError    // 1.34
+];
--- a/compatibility-tests/without-backtrace/Cargo.toml
+++ b/compatibility-tests/without-backtrace/Cargo.toml
@@ -5,4 +5,4 @@ authors = ["Jake Goulding <jake.goulding@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-snafu = { path = "../..", default-features = false, features = [] }
+snafu = { path = "../..", default-features = false, features = ["std"] }
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -1298,7 +1298,7 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
         let backtrace_field = match *backtrace_field {
             Some(ref field) => {
                 let name = &field.name;
-                quote! { #name: std::default::Default::default(), }
+                quote! { #name: core::default::Default::default(), }
             }
             None => quote! {},
         };
@@ -1308,7 +1308,7 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
             .zip(user_fields)
             .map(|(gen_ty, f)| {
                 let Field { ref ty, .. } = *f;
-                quote! { #gen_ty: std::convert::Into<#ty> }
+                quote! { #gen_ty: core::convert::Into<#ty> }
             })
             .chain(self.0.provided_where_clauses())
             .collect();
@@ -1318,16 +1318,16 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
                 impl<#(#generic_names,)*> #selector_name
                 {
                     #[doc = "Consume the selector and return a `Result` with the associated error"]
-                    #visibility fn fail<#(#original_generics_without_defaults,)* __T>(self) -> std::result::Result<__T, #parameterized_enum_name>
+                    #visibility fn fail<#(#original_generics_without_defaults,)* __T>(self) -> core::result::Result<__T, #parameterized_enum_name>
                     where
                         #(#where_clauses),*
                     {
                         let Self { #(#names),* } = self;
                         let error = #enum_name::#variant_name {
                             #backtrace_field
-                            #( #names: std::convert::Into::into(#names) ),*
+                            #( #names: core::convert::Into::into(#names) ),*
                         };
-                        std::result::Result::Err(error)
+                        core::result::Result::Err(error)
                     }
                 }
             }
@@ -1367,7 +1367,7 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
             quote! {
                 impl#generics_list snafu::IntoError<#parameterized_enum_name> for #selector_name
                 where
-                    #parameterized_enum_name: std::error::Error + snafu::ErrorCompat,
+                    #parameterized_enum_name: snafu::Error + snafu::ErrorCompat,
                     #(#where_clauses),*
                 {
                     type Source = #source_ty;
@@ -1453,11 +1453,11 @@ impl<'a> quote::ToTokens for DisplayImpl<'a> {
 
         stream.extend({
             quote! {
-                impl<#(#original_generics),*> std::fmt::Display for #parameterized_enum_name
+                impl<#(#original_generics),*> core::fmt::Display for #parameterized_enum_name
                 where
                     #(#where_clauses),*
                 {
-                    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
                         #[allow(unused_variables)]
                         match *self {
                             #(#variants_to_display)*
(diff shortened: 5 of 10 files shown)


Cargo.toml
```diff
@@ -23,22 +23,25 @@ exclude = [
...
 # Add extension traits for the futures 0.1 crate
-"futures-01" = ["futures01"]
+"futures-01" = ["std", "futures01"]
 
 # New methods on `Error`; re-export of proc-macro
 rust_1_30 = ["snafu-derive/rust_1_30", "doc-comment"]
 
 # The standard library's implementation of futures
-"unstable-futures" = ["futures-core-preview", "pin-project"]
+"unstable-futures" = ["std", "futures-core-preview", "pin-project"]
```
shepmaster: `Future` is in core; why is this needed?

Cargo.toml
```diff
@@ -23,22 +23,25 @@ exclude = [
...
-backtraces = ["snafu-derive/backtraces", "backtrace"]
+backtraces = ["std", "snafu-derive/backtraces", "backtrace"]
 
 # Add conversion to `backtrace::Backtrace`
-backtrace-crate = ["backtraces"]
+backtrace-crate = ["std", "backtraces"]
 
 # Add extension traits for the futures 0.1 crate
-"futures-01" = ["futures01"]
+"futures-01" = ["std", "futures01"]
```
shepmaster: Futures 0.1 supports no-std; why do we need to enforce it?

Cargo.toml
```diff
@@ -23,22 +23,25 @@ exclude = [
...
+# Uses the `std::error::Error` trait instead of its own
+std = []
 
 # Adds the backtrace type
-backtraces = ["snafu-derive/backtraces", "backtrace"]
+backtraces = ["std", "snafu-derive/backtraces", "backtrace"]
 
 # Add conversion to `backtrace::Backtrace`
-backtrace-crate = ["backtraces"]
+backtrace-crate = ["std", "backtraces"]
```
shepmaster: Since this implies `backtraces`, I'd prefer to just have `std` there.

Cargo.toml
```diff
@@ -23,22 +23,25 @@ exclude = [
...
 [features]
-default = ["rust_1_30", "backtraces"]
+default = ["rust_1_30", "backtraces", "std"]
+
+# Uses the `std::error::Error` trait instead of its own
+std = []
 
 # Adds the backtrace type
-backtraces = ["snafu-derive/backtraces", "backtrace"]
+backtraces = ["std", "snafu-derive/backtraces", "backtrace"]
```
shepmaster: The backtrace crate has a `std` flag; why do we need to enforce it?

src/no_std_error.rs
```diff
@@ -0,0 +1,199 @@
...
+///
+/// The [`source`] method is generally used when errors cross "abstraction
+/// boundaries". If one module must report an error that is caused by an error
+/// from a lower-level module, it can allow access to that error via the
+/// [`source`] method. This makes it possible for the high-level module to
+/// provide its own errors while also revealing some of the implementation for
+/// debugging via [`source`] chains.
+///
+/// [`source`]: trait.Error.html#method.source
+pub trait Error: Debug + Display {
```
shepmaster: I'm very hesitant to copy the standard libraries `Error` like this, especially the implementations on arbitrary types. One thing I've been negative about is how failure created a new trait and then told everyone that they needed to switch to it to be compatible. This felt like it was a bad "team player" in the broader ecosystem.

I suppose we'd quickly run into coherence problems if we make this a marker trait and provided a blanket implementation?

```rust
pub trait Error: Debug + Display {}
impl<T: Debug + Display> Error for T {}
```

Another avenue is to completely drop `Error` and just implement `Debug + Display`; any idea what that world might look like?

Cargo.toml
CryZe: I've not been looking at which other features no_std could support yet. So for now I simply marked all of them as std only. But yeah unstable-futures could probably work with no_std.

src/no_std_error.rs
CryZe: I've considered that when I initially tried implementing this, but you quickly lose out on most of snafu's feature such as the ResultExt and OptionExt traits. We'd at least need some trait that we implement in addition to Display and Debug so we can somehow make ResultExt and OptionExt work too. Also not having source chaining is probably pretty bad too. So that's why in my initial prototyping it evolved from not having an Error trait at all to eventually just a Copy of the full thing (although I did consider dropping the deprecated methods).

CryZe: The Error trait introduced here is not really meant to be used anywhere really. In livesplit-core where I've been using this no-std snafu branch the Error trait never is publicly visible to anyone. It is mostly just a convenience trait to not lose out on any of the features of snafu.

CryZe: Actually the blanket implementation Error trait could maybe work. I'll see if it causes any problems.

Cargo.toml
CryZe: When you want the backtrace crate, you have std and thus the std feature should be active such that you get the std::error::Error implementations and not the ones for the core Error trait we introduce here.

src/no_std_error.rs
CryZe: Although I don't think there even is any harm to having a copy of the Error trait in snafu. You can still treat core as if it didn't have an Error trait at all. But if you want to use an Error trait snafu optionally has one available for you to use. And if someone activates the std feature (either directly through a feature of your crate or indirectly), your code automatically uses the std Error trait and there is no "bad team playing" going on.

CryZe: Although one problem I can see is that errors from other core crates might not implement an Error trait at all on core, so you can't use ResultExt and co. regardless.

Cargo.toml
CryZe: Done

CryZe:
I don't quite understand what's wrong with the stable and nightly build. These modules compile just fine on stable and nightly.

CryZe:
1.18 can't be supported unless we also modify the macro such that it conditionally emits paths to std instead of core. However this PR is breaking for everyone anyway with `default-features = false`, so depending on how we want to handle this, this may need to be part of a snafu 0.5 instead and 1.18 support could be removed at the same time, so we don't need to special case everything in the macro too.

CryZe:
Okay, so here's a proposal: We conservatively go for `doc(hidden)` such that there is no publicly visible Error trait, but snafu still retains all of its features to reduce friction when a user is trying to support both std and no_std. At a later point the trait can either transparently be switched out by the real core::error::Error or be made publicly visible without causing any breaking changes.

CryZe:
This is basically done from my side. The CI server seems to not be responding, but it should turn green eventually and everything is implemented.

Cargo.toml
CryZe: Oh wait, you meant it works with no_std? I guess I'll try to get that working.

CryZe: Alright, I implemented the backtrace support. I was not quite able to test it properly though.

CryZe: Nvm, this doesn't build at all. You can't capture any backtraces with no_std, so I'd argue that we can't really support them until that's a thing in the backtrace crate.

CryZe:
What's the status of this?

CryZe:
I rebased this onto the latest changes to resolve the conflicts. This is ready for review again :)

Cargo.toml
shepmaster: Sounds good, I guess that's just a placeholder functionality in their code.

compatibility-tests/v1_18/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![cfg(test)]
 
+extern crate core;
```
shepmaster: I saw you said

> it can be supported if we put extern crate core; in the test, but that is a breaking change for everyone using 1.18

I wanted to see which versions would be affected, so I removed this line and built in 1.18, but I didn't get an error; any idea what might have changed or that I might be doing wrong?

CryZe: It's actually Rust 2015 that's affected. The proc macro targets core, but that's not in the prelude in Rust 2015.

src/no_std_error.rs
shepmaster: It feels like we will need to release *something* and get feedback from people as to the usefulness. I think it's best we call the no-std support / exact implementation experimental.

For example, one thing that concerns me is what happens if one crate using no-std SNAFU exports an error type. Will other crates be able to make use of it, or will they need to wrap it in that newtype?

src/no_std_error.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// fn main() {
+    ///     match get_super_error() {
+    ///         Err(e) => {
+    ///             println!("Error: {}", e.description());
+    ///             println!("Caused by: {}", e.source().unwrap());
+    ///         }
+    ///         _ => println!("No error"),
+    ///     }
+    /// }
+    /// ```
```
shepmaster: I get that this is copied from the standard library, but since we do our best to hide the documentation, I think I'd feel better stripping this to a bare minimum. That way it can't fall out of date with the standard library and people might be dissuaded if they do stumble on the docs; something like:

```rust
/// This is a placeholder {trait, method}
```

We could also consider just adding `#![allow(missing_docs)]` to this module.

src/no_std_error.rs
CryZe: I'm not sure I understand what you mean with that. Why would they need to wrap it in a newtype? If the error type is from snafu in a no_std context, then there simply is no Error trait, just like usual. Or are you saying what the story is with two crates using snafu in a no_std context? That should work seamlessly too, unless they are using different snafu versions.

src/no_std_error.rs
CryZe: Alright yeah I can get rid of it.

src/no_std_error.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
```
shepmaster: These `dyn`s won't work in Rust 1.18

CryZe: ah, right, I'll fix that.

src/no_std_error.rs
```diff
@@ -0,0 +1,199 @@
...
+        None
+    }
+}
+
+impl Error for core::str::ParseBoolError {
```
shepmaster: Perhaps it would be better to just do the bare minimum for all core errors, not even the `description` method. Then we could even do a macro to trim it down to the max:

```rust
some_name![
    core::str::ParseBoolError,
    core::str::Utf8Error,
    core::num::ParseIntError,
];
```

Then we can sort these alphabetically to track changes.

src/no_std_error.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn description(&self) -> &str {
+        "already borrowed"
+    }
+}
+impl Error for core::char::CharTryFromError {
+    fn description(&self) -> &str {
+        "converted integer out of range for `char`"
+    }
+}
+impl Error for core::char::ParseCharError {}
```
shepmaster: Speaking of "tracking changes", what would we do when (for example) Rust 1.40 adds a new `core::foo::FooError`? That type won't exist in Rust 1.39, so we can't add it here without breaking backwards compatibility. If we don't add it, then people won't be able to use the fancy features, as you mentioned.

CryZe: That is actually already kind of the case with some of the Errors here, they aren't stable in 1.18. So maybe we can do rustc version detection via a build.rs? That would get rid of the 1.30 feature too.

CryZe:
Good news, looks like the Error trait is moving into the alloc crate: https://github.com/rust-lang/rust/pull/64017

roblabla:
Bad news, the error trait can't move into the alloc trait because of RFC 2504

CryZe:
Yeah while not the best news, I still think the approach in this PR works fairly well and can easily be extended to alloc::Error whenever someone figures out how to make it compatible with the backtraces. I think you want the core::Error introduced here anyway as a fallback for when you don't even want alloc.

roblabla:
With fehler released, and failure still around, I was wondering if maybe it would be a good idea to move the `Error` trait in a separate crate, tentatively called `core-error`, that would be co-maintained by the different stakeholders of error libraries?

My hope is that it would help with the interoperability, and get no_std libraries that don't want to buy into a specific error handling crate (snafu/failure...) to still interoperate with them.

CC https://github.com/withoutboats/fehler/issues/12#issuecomment-533795644

CryZe:
Yeah I was considering that when we had the discussion about the error trait here. There already is a crate, but it seems barely used / maintained. But maybe it can be "revived"? I should definitely have a std feature and auto upgrade to std's Error if it exists, so you automatically get upgraded to the proper Error trait for the specific use case.

roblabla:
Do you have a link to said crate? A quick google didn't turn up anything.

CryZe:
Here: https://crates.io/crates/error_core

src/no_std_error.rs
```diff
@@ -0,0 +1,45 @@
...
+impl_error![
+    core::str::ParseBoolError,    // 1.0
+    core::str::Utf8Error,         // 1.0
+    core::num::ParseIntError,     // 1.0
+    core::num::ParseFloatError,   // 1.0
+    core::char::DecodeUtf16Error, // 1.9
+    core::fmt::Error,             // 1.11
+    core::cell::BorrowMutError,   // 1.13
+    core::cell::BorrowError,      // 1.13
+    core::char::ParseCharError    // 1.20
```
shepmaster: Since we bumped our minimum version to 1.31, are there any errors missing from this list?

shepmaster:
@roblabla I think that's a great idea! I'd be much happier knowing that we weren't making some completely custom trait that was tied to this crate. Ideally, some magical solution would come from libcore, but until that happens, a shared community crate seems like a great step.

Since this PR is likely to be resolved before that crate can completely be usable for our cases, would you mind opening a new issue to track the idea?

src/no_std_error.rs
CryZe: No that I'm aware of. When I created the PR these were all the Error impls in std. These used to be three macro calls where one had the rust_1_30 feature check, but I condensed that into the unconditional macro call, so I don't think I missed any (other than potential ones in 1.38+ or so).

.cirrus.yml
```diff
@@ -27,6 +27,22 @@ stable_test_task:
...
+    cpu: 1
+    memory: 2Gi
+  cargo_cache:
+    folder: $CARGO_HOME/registry
+    fingerprint_script: cat Cargo.toml
+  setup_script:
+    - rustup target add thumbv6m-none-eabi
+  primary_test_script:
+    - rustc --version
+    - cargo build --no-default-features --target thumbv6m-none-eabi
```
shepmaster: Completely unrelated, but thanks for showing me how to actually make some kind of test for a no-std crate. I had some [*really* nasty](https://github.com/shepmaster/twox-hash/blob/5c9a4690f576a20280a9ecbc4cbdc3ab2fff26e2/ci/check-no-std.sh#L25-L29) hacks to try and do this before.

shepmaster:
Thanks!

PR #141, opened 2019-08-07
Replace `backtrace(delegate)` with `backtrace` on a source field

### LINKED ISSUE #125: Replacing `backtrace(delegate)` with `backtrace` on a source field
In #109, an idea was raised - is `backtrace(delegate)` needed as a separate attribute?

Its purpose is to signal that another error type will handle the backtrace.  (See https://docs.rs/snafu/0.4.3/snafu/guide/attributes/index.html#controlling-backtraces).  For this to work, that other type, where you add the `backtrace(delegate)` attribute, must be a source field.  And if it's a source field, it can't *directly* be a backtrace field; sources must be other error types, and backtraces must be `Backtrace`.

So, can we simplify and say that backtrace delegation happens by marking a source field with a `backtrace` attribute?

Example with a typical "source" field:

```rust
#[derive(Debug, Snafu)]
enum Error {
    MyError {
        #[snafu(backtrace)]
        source: another::Error,
    },
}
```

Example with a `source` attribute, required to handle another field name:
```rust
#[derive(Debug, Snafu)]
enum Error {
    MyError {
        #[snafu(source, backtrace)]
        my_source: another::Error,
    },
}
```

Source transformation shouldn't be affected; and with #124 (`source(from)` implying `source(true)`) even the most complicated case seems relatively simple:

```rust
#[derive(Debug, Snafu)]
enum Error {
    MyError {
        #[snafu(backtrace)]
        #[snafu(source(from(another::Error, Box::new)))]
        my_source: Box<another::Error>,
    },
}
```

tjkirch:
Fixes #125, see it for background.

As I was playing with some options, I came to believe that the clearest representation of this change is adding a flag to `SourceField` that represents whether the source field is a backtrace delegate.  We don't need the `backtrace_delegate` field in `VariantInfo` with its `Option<Field>`; all delegates are source fields, and we already have the source field information.

I was then able to simplify `parse_snafu_enum` by removing `backtrace_delegates` and `backtrace_delegate_location` - we can detect whether it's a delegate just by seeing if we got a backtrace attribute on a source field, both of which were already tracked, so we don't need separate tracking methods.

I added a friendly error if we do see `backtrace(delegate)`, and updated the tests and guide.  Updated compile-fail tests show the (even better?) errors for duplication.

No idea why rustfmt decided to rewrap some unchanged paragraphs, sorry :(

I'm interested in any feedback on this approach of moving knowledge of backtrace delegation to `SourceField`, and of course any other general feedback.

### DIFF
--- a/compatibility-tests/compile-fail/tests/ui/duplication-backtrace-delegate.stderr
+++ b/compatibility-tests/compile-fail/tests/ui/duplication-backtrace-delegate.stderr
@@ -1,11 +1,12 @@
-error: Cannot have `backtrace` and `backtrace(delegate)` fields in the same error variant
+error: Cannot have `backtrace` field and `backtrace` attribute on a source field in the same error variant
+  --> $DIR/duplication-backtrace-delegate.rs:10:9
+   |
+10 | /         #[snafu(backtrace)]
+11 | |         source: String,
+   | |______________________^
+
+error: Cannot have `backtrace` field and `backtrace` attribute on a source field in the same error variant
  --> $DIR/duplication-backtrace-delegate.rs:7:9
   |
 7 |         backtrace: Backtrace,
   |         ^^^^^^^^^^^^^^^^^^^^
-
-error: Cannot have `backtrace` and `backtrace(delegate)` fields in the same error variant
-  --> $DIR/duplication-backtrace-delegate.rs:10:17
-   |
-10 |         #[snafu(backtrace(delegate))]
-   |                 ^^^^^^^^^^^^^^^^^^^
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -44,7 +44,6 @@ struct VariantInfo {
     name: syn::Ident,
     source_field: Option<SourceField>,
     backtrace_field: Option<Field>,
-    backtrace_delegate: Option<Field>,
     user_fields: Vec<Field>,
     display_format: Option<UserInput>,
     doc_comment: String,
@@ -72,6 +71,7 @@ impl Field {
 struct SourceField {
     name: syn::Ident,
     transformation: Transformation,
+    backtrace_delegate: bool,
 }
 
 impl SourceField {
@@ -474,8 +474,6 @@ fn parse_snafu_enum(
             let mut user_fields = Vec::new();
             let mut source_fields = AtMostOne::new("source", "within an error variant");
             let mut backtrace_fields = AtMostOne::new("backtrace", "within an error variant");
-            let mut backtrace_delegates =
-                AtMostOne::new("backtrace(delegate)", "within an error variant");
 
             for syn_field in fields {
                 let span = syn_field.span();
@@ -504,10 +502,6 @@ fn parse_snafu_enum(
                 let mut source_opt_out = false;
                 let mut backtrace_opt_out = false;
 
-                // Keep track of the location of the backtrace delegate field, if any; duplication
-                // is already checked because backtrace delegates are counted in backtrace_attrs.
-                let mut backtrace_delegate_location = None;
-
                 for attr in attributes_from_syn(syn_field.attrs.clone())? {
                     match attr {
                         SnafuAttribute::Source(tts, ss) => {
@@ -517,7 +511,8 @@ fn parse_snafu_enum(
                                     Source::Flag(v) => {
                                         // If we've seen a `source(from)` then there will be a
                                         // `Some` value in `source_attrs`.
-                                        let seen_source_from = source_attrs.iter()
+                                        let seen_source_from = source_attrs
+                                            .iter()
                                             .map(|&(ref val, ref _location)| val)
                                             .any(Option::is_some);
                                         if !v && seen_source_from {
@@ -550,19 +545,13 @@ fn parse_snafu_enum(
                                 }
                             }
                         }
-                        SnafuAttribute::Backtrace(tts, b) => match b {
-                            Backtrace::Flag(v) => {
-                                if v {
-                                    backtrace_attrs.add((), tts);
-                                } else {
-                                    backtrace_opt_out = true;
-                                }
-                            },
-                            Backtrace::Delegate => {
-                                backtrace_attrs.add((), tts.clone());
-                                backtrace_delegate_location = Some(tts);
+                        SnafuAttribute::Backtrace(tts, v) => {
+                            if v {
+                                backtrace_attrs.add((), tts);
+                            } else {
+                                backtrace_opt_out = true;
                             }
-                        },
+                        }
                         SnafuAttribute::Visibility(tts, ..) => {
                             errors.add(
                                 tts,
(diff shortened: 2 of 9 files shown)


compatibility-tests/compile-fail/tests/ui/duplication-backtrace-delegate.stderr
```diff
@@ -1,11 +1,12 @@
-error: Cannot have `backtrace` and `backtrace(delegate)` fields in the same error variant
+error: Cannot have `backtrace` field and `backtrace` attribute on a source field in the same error variant
+  --> $DIR/duplication-backtrace-delegate.rs:10:9
+   |
+10 | /         #[snafu(backtrace)]
+11 | |         source: String,
+   | |______________________^
```
shepmaster: This highlight seems a bit odd; are we spanning more than we should?

snafu-derive/src/lib.rs
```diff
@@ -919,10 +904,9 @@ impl syn::parse::Parse for Source {
     }
 }
 
-enum Backtrace {
-    Flag(bool),
-    Delegate,
-}
+// Having a Backtrace newtype and implementing Parse gives us a way to handle careful parsing
+// errors outside of `impl Parse for SnafuAttribute`
```
shepmaster: I really appreciate that you've been adding documentation; I tend to be pretty lazy for the "internal" docs.

snafu-derive/src/lib.rs
```diff
@@ -932,17 +916,17 @@ impl syn::parse::Parse for Backtrace {
...
         } else if lookahead.peek(Ident) {
             let name: Ident = input.parse()?;
 
             if name == "delegate" {
-                Ok(Backtrace::Delegate)
-            } else {
                 Err(syn::Error::new(
                     name.span(),
-                    "expected `true`, `false`, or `delegate`",
+                    "`backtrace(delegate)` has been removed; use `backtrace` on a source field",
```
shepmaster: I like this kind of helpful text, but I also would only want to keep it for "so long". I wonder if there's a way to indicate that this code can be removed at some future point (that I don't know what it would be)

tjkirch: I agree, but I didn't have a great idea other than perhaps adding a comment.  "Removal message added 2019-08-07, delete when people have forgotten about delegate" or something :)

Or perhaps as one of the line items for a 1.0 release?

tjkirch:
> What happen if we add `backtrace(false)` on a `source` field? I could either see it being allowed (it' s just meaningless) or an error (you probably didn't mean that).

Right now, it would allow it.  It seems to me like a potential confusion, but not strictly wrong - something I'd warn about when we get the ability to warn.  Do you agree?

compatibility-tests/compile-fail/tests/ui/duplication-backtrace-delegate.stderr
tjkirch: Ah, yes, I found it.  It's because for fields named "source" or "backtrace" I was getting the location using `syn_field`, which represents the entire thing with attributes.

In this specific example, I kind of like the longer span, because it makes the "backtrace attribute on a source field" idea clear.  However, it'd be less clear if there were more attributes on the field - they'd also be highlighted.

I'm not sure the best thing to point at, though, if we wanted to restrict it.  "backtrace attribute on a source field" has those two inner concepts that are both important.  I'd guess the attribute is more important.  But, then, I'm not sure how to get the location of the backtrace attribute on the source field at the time when we're checking whether we saw a source attribute.  Either I'd store the backtrace location as another awkward piece of data in the source attribute tracking, or dig back through the attributes on the field, or...

I think the larger span is actually not a bad compromise...

tjkirch:
Marked as "ready for review" because it seemed like the general approach was OK, and there weren't any hard blockers in the discussion above, though I'm happy to change course as the discussion proceeds!

tjkirch:
This push just rebases onto 0.4.4.

compatibility-tests/compile-fail/tests/ui/duplication-backtrace-delegate.stderr
shepmaster: I don't think we have the full procedural macro power we need with the current version of Rust. The compiler proper has the ability to do multiple spans; I'd love it if it could be

```
cannot have backtrace attribute on source field when there is already a backtrace field
NN #[snafu(backtrace)]
   ^~~~~~~~~~~~~ attribute declared here
NN source: Box<Error>,
    ^~~~~~~~~ field declared here
```

snafu-derive/src/lib.rs
shepmaster: It's probably not important either way. Let's just leave it until it gets in the way and can then evaluate.

compatibility-tests/compile-fail/tests/ui/duplication-backtrace-delegate.stderr
shepmaster: If that's not possible, then I'm OK with the longer span. Reporting the error is the backwards compatibility issue, improving the span (if possible) should be able to happen later.

shepmaster:
> just rebases onto 0.4.4.

It shouldn't matter here as 0.4.4 and master have been merged and are effectively the same, but you'll want to rebase on master in the future (if there's even any more patches)

tjkirch:
> It shouldn't matter here as 0.4.4 and master have been merged and are effectively the same, but you'll want to rebase on master in the future (if there's even any more patches)

Sorry, that's what I meant; I rebased onto master, and the change in doing so was including 0.4.4.  :)

tjkirch:
> Do we handle putting `source` and `backtrace` on the same field?

Yep; it's a bit simpler with `backtrace_delegate` as a bool:

https://github.com/shepmaster/snafu/pull/141/files#diff-89e1fad4d2637dae0904dce6aba08394R611-R613

I realized that I forgot to add a test for both using attributes, which I'll do.  (I only added a test for a field named "source" with a `backtrace` attribute; this would test the same code path now, but maybe not in the future.)

I also realized I hadn't tried exactly `#[snafu(source, backtrace)]` and unfortunately it's unhappy, saying that it expected parentheses after `source`.  :(  I expect anything using `MyParens` would have that problem.  I'll make a new issue and try to figure that out.

compatibility-tests/compile-fail/tests/ui/duplication-backtrace-delegate.stderr
tjkirch: There was https://github.com/dtolnay/syn/issues/558 requesting errors with multiple spans in syn, which was closed in preference to it being built in another library.  It seems the [proc-macro Diagnostic](https://doc.rust-lang.org/stable/proc_macro/struct.Diagnostic.html) API might be the way of the future, once it's implemented and stabilized in https://github.com/rust-lang/rust/issues/54140.  I agree that it doesn't seem like we have a convenient way currently.

tjkirch:
I just pushed a new test showing `source` and `backtrace` working together on one field.  It does use `source(true)` rather than `source` because of #146, which I plan to address separately in that issue.

tjkirch:
(Missed rustfmt, sorry!)

tjkirch:
Interesting.  The Rust 1.18 compatibility test passed on my machine because I already had a Cargo.lock with rustc-demangle 0.1.15, but a fresh build pulls in rustc-demangle 0.1.16 (released yesterday) which causes this error:

```
error: no method named `is_control` found for type `char` in the current scope
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rustc-demangle-0.1.16/src/legacy.rs:168:43
    |
168 |                                     if !c.is_control() {
    |                                           ^^^^^^^^^^
```

Here's that failing call in context with the version diff: https://github.com/alexcrichton/rustc-demangle/compare/0.1.15...0.1.16#diff-f528d4fa5ce70b178b63beb6758dd35cR168

I'm trying to figure out whether `core::char::is_control` wasn't in Rust 1.18... it seems like `std::char::is_control` was...

tjkirch:
I went through the Rust release note history, and I've been digging through the Rust commit history, and can't see anything significant changing about `is_control` since 1.18.  It seems to have existed in `core` since 1.0.  It's definitely moved around (`core::unicode::char`, `core::unicode`, `libstd_unicode`, etc.) but I don't know why that would break this.

@shepmaster - I'm assuming you've dealt with issues where dependencies (direct or transitive) haven't supported 1.18 - what do you typically do?  Is that the dependency list in `compatibility-tests/v1_18/Cargo.toml`?  I didn't realize that could force versions of transitive dependencies, too, so that's handy.  But wouldn't all customers using 1.18 need to know to do the same?

Is that appropriate for this case, or do you have any idea about rustc-demangle and `is_control`?

shepmaster:
> I don't know why that would break this.

Yeah, that’s super weird and I don’t see anything obvious. However, I can reproduce the problem locally in a brand new project with no dependencies. 

> Is that the dependency list in compatibility-tests/v1_18/Cargo.toml? 

Yep!

> I didn't realize that could force versions of transitive dependencies, too, so that's handy. 

I think it’s mostly kind of lucky right now, I think. Cargo tries to have a unified version, so it is kind of indirect. If we specified an incompatible version, wed get two different versions and one would fail to build. Then we’d have to move back to the old `cargo update` invocations. 

>But wouldn't all customers using 1.18 need to know to do the same?

Kind of. Hopefully they’d already have these crates before adding SNAFU. If not, they are likely used to manually downgrading versions.

shepmaster:
I’ve merged an update that should fix. 1.18; rebasing should grab it.

tjkirch:
Thanks!  Rebased to include that change.

PR #157, opened 2019-08-17
Make "backtraces" feature opt-in

lucab:
This moves the "backtraces" feature out of the default set. It also
links the `Backtrace` implementation to the feature, so that the
public type always exists but it does not carry any backtrace content
unless the consumer explicitely enabled that.

Closes: https://github.com/shepmaster/snafu/issues/113


lucab:
Rebased to latest master (and fixed a bunch of conflicts), CI is green.

shepmaster:
My plan is to take this PR and PR #173 and ensure that they play well together locally before I merge either, then merge them separately. 

Since our goal is to support the standard-library backtrace and that will be always available in some future Rust version, that's the general setup I want to mimic. I think this PR sets us up in that direction, but I don't know if it will be the final goal.

Let's assume that standard-library backtraces stabilize in Rust 1.50. My plan is that this library would grow a `rust-1.50` feature flag which enables them, re-exporting it as `snafu::Backtrace`. Without that feature flag, `Backtrace` would be the inert version here. An *alternate* (and conflicting!) feature flag would use `backtrace::Backtrace` (potentially merging the `backtrace-crate` feature in at that point too).

shepmaster:
I'll take care of addressing these conflicts, as part of that local testing!

shepmaster:
Thanks!

PR #161, opened 2019-08-26
Update user-facing release information

shepmaster:


shepmaster:
I'd love a review, but please don't merge this; I want to make sure the dates are correct for when I actually push it ;-)

shepmaster:
The main differences from your suggested content are:

- added `#[snafu...]` to the changelog entries.
- rephrased the changelog for a consistent style (not as worried about that for the upgrading guide as it's chunkier to read)
- reordered the upgrading guide in "most likely to experience" order

CryZe:
I just fixed the review findings for the no-std stuff, so if you want, you could experimentally add it to 0.5. It's fine if it doesn't make it into 0.5 though, my code is blocked by many other crates not being no-std yet anyway.

PR #171, opened 2019-09-11
Document context selectors and mention semver breakage when exporting them

### LINKED ISSUE #168: Generated context selectors using public visibility cause failures in a crate with deny(missing_docs)
I ran into a rough edge while porting some of my code to Snafu today. The crate has a `!#[deny(missing_docs)]` attribute and exports a snafu error type, I'd like to also export the generated context selectors so I'm using `#[snafu(visibility = "pub")]` as such:

```rust
#[derive(Debug, Snafu)]
#[snafu(visibility = "pub")]
/// ...
pub enum MyError {
    #[snafu(display(...))]
    /// HTTP error.
    Http { ... },
    #[snafu(display(...))]
    /// Network error.
    Network { ... },
    #[snafu(display(...))]
    /// Deserialization error.
    Serde { ... }
}
```

As the generated selectors don't have documentation, I get a compile error, and since the selector types are generated by the macro, I can't really add an `#[allow(missing_docs)]` annotation or add the documentation myself (at least, I couldn't find any documented way). The workaround I use is to define my snafu error type in a module with `!#[allow(missing_docs)]` at the top and `pub use x::*` from the parent module, turning the previous code into:

```rust
mod snafu_exports {
    #![allow(missing_docs)]

    #[derive(Debug, Snafu)]
    #[snafu(visibility = "pub")]
    /// ...
    pub enum MyError {
        #[snafu(display(...))]
        /// HTTP error.
        Http { ... },
        #[snafu(display(...))]
        /// Network error.
        Network { ... },
        #[snafu(display(...))]
        /// Deserialization error.
        Serde { ... }
}
pub use snafu_exports::*;
```

Which seems to work, but not having to use this workaround would be better. The simplest solutions I see are:

* Generate some default documentation (something like "Snafu context selector for the X error variant" ?)
* Provide a way to write the documentation (but would that be very useful ?)
* Annotate/provide a way to annotate the generated code with `#[allow(missing_docs)]`.

### LINKED ISSUE #170: Enhance documentation about visibility to warn about semver concerns when exporting from the crate
We don't really make any guarantees that the code we generate (e.g. context selectors) is API-stable. If someone chooses to export them from their crate, then their semver requirements are tied to SNAFU's whims.

Robzz:
This PR generates some default documentation for SNAFU context selector types and adds a mention of possible semver breakage in the user guide, in the section discussing the visibility attribute.

I'm opening this already, so that we can discuss and improve the wording of the documentation enhancements if needed. I'll deal with generating the documentation when I have the time to dive into `snafu-derive`, hopefully later this week.

Closes #168  
Closes #170

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -19,6 +19,8 @@ stable_test_task:
     - cargo test --manifest-path compatibility-tests/without-backtrace/Cargo.toml
   futures_0.1_test_script:
     - cargo test --manifest-path compatibility-tests/futures-0.1/Cargo.toml
+  context_selectors_have_documentation_test_script:
+    - cargo test --manifest-path compatibility-tests/context-selectors-have-documentation/Cargo.toml
   compile_fail_test_script:
     - cargo test --manifest-path compatibility-tests/compile-fail/Cargo.toml
   lint_script:
--- a/compatibility-tests/context-selectors-have-documentation/Cargo.toml
+++ b/compatibility-tests/context-selectors-have-documentation/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "context-selectors-have-documentation"
+version = "0.1.0"
+authors = ["Jake Goulding <jake.goulding@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+snafu = { path = "../.." }
--- a/compatibility-tests/context-selectors-have-documentation/src/lib.rs
+++ b/compatibility-tests/context-selectors-have-documentation/src/lib.rs
@@ -1,0 +1,13 @@
+#![deny(missing_docs)]
+
+//! Crate docs
+
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
+/// Enum docs
+pub enum Error {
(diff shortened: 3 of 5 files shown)


shepmaster:
> [WIP] 

FYI, GitHub has [draft pull requests not](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

Robzz:
> FYI, GitHub has [draft pull requests now](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

Didn't know that, thanks!

I just added the selectors documentation. As predicted, I also had to document the selector fields, but I couldn't think of a helpful default documentation so I just added `#allow[missing_docs]` to those. I also had to document the `fail` method that's generated for some selectors.

Let me know if you'd like any changes.

Robzz:
Apparently the CI was unhappy about the formatting, I ran rustfmt and it now passes on rust stable. It still fails on rust 1.18 and 1.30 but it looks unrelated to this PR.

shepmaster:
> It still fails on rust 1.18 and 1.30

We are actually right in the middle of some reorganizing and breaking changes, so I rebased you on master where we now support only Rust 1.31+. I made some small tweaks to the language and added a test to ensure that the docs are generated. I force-pushed all that to your branch.

Let me know if the changes look good to you!

Robzz:
It still builds with my usecase so yeah, all good!

Issue #172, opened 2019-09-11
Add support for the standard-library backtrace type


shepmaster:
Hot off the presses: https://github.com/rust-lang/rust/pull/64154, might need to wait a day before it's in nightly.

shepmaster:
Closed by #173

Issue #84, opened 2019-05-18
Investigate ways to avoid performance impacts of backtraces


shepmaster:
Right now, if you create a backtrace-enhanced error but then handle it, you've paid for the cost to walk the stack and resolve the symbols for no reason. This *may* cause performance issues.

Two ideas I have:

1. Use an environment variable to prevent backtraces from being generated

    Look for something like `RUST_BACKTRACE=1` — if it's not provided the backtrace becomes a no-op. Great for performance but sucks when you don't have it enabled when you need it.

    May want to make this a separate type so that people can opt-in to either behavior.

1. Don't resolve the symbols until display time

    I believe the backtrace crate supports this. This would reduce the cost of generating the backtrace. This may be a "free" win and we should do it unconditionally.

shepmaster:
It would be nice to have some quantitative idea of how expensive generating a backtrace is, as a baseline.

yaahc:
This is exactly how `failure` handles backtraces and is one of the features I'm really missing from it, you could talk to boats about bringing their backtrace logic into snafu.

One approach I was thinking might make sense is to allow users to have the type of their backtrace type in their error be `Option<Backtrace>` much like how structopt allows you to make a flag optional by wrapping it in an `Option`, that way it's configurable on a per usage basis whether or not you want to unconditionally collect backtraces or base it on the `RUST_BACKTRACE`.

PR #173, opened 2019-09-21
Add support for standard library std::backtrace::Backtrace

ccfb3ee765a58cae:
See #172. This gives a proper `.backtrace()` to snafu errors, using nightly's `#![feature(backtrace)]`.

When enabled, this re-exports `std::backtrace::Backtrace` as `snafu::Backtrace` and changes `snafu::ErrorCompat::backtrace` such that it returns a std library backtrace instead of one from the `backtrace` crate.

May be a bit messy, as I haven't contributed to this project before. There are some considerations that need to be decided (see git commit comments) before this is merged, so I'm leaving it as a draft.

### DIFF
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -1560,6 +1560,16 @@ impl<'a> quote::ToTokens for ErrorImpl<'a> {
             }
         };
 
+        let std_backtrace_fn = if cfg!(feature = "unstable-backtraces-impl-std") {
+            quote! {
+                fn backtrace(&self) -> Option<&std::backtrace::Backtrace> {
+                    snafu::ErrorCompat::backtrace(self)
+                }
+            }
+        } else {
+            quote! {}
+        };
+
         stream.extend({
             quote! {
                 impl<#(#original_generics),*> snafu::Error for #parameterized_enum_name
@@ -1570,6 +1580,7 @@ impl<'a> quote::ToTokens for ErrorImpl<'a> {
                     #description_fn
                     #cause_fn
                     #source_fn
+                    #std_backtrace_fn
                 }
             }
         })
@@ -1690,6 +1701,16 @@ impl StructInfo {
             }
         };
 
+        let std_backtrace_fn = if cfg!(feature = "unstable-backtraces-impl-std") {
+            quote! {
+                fn backtrace(&self) -> Option<&std::backtrace::Backtrace> {
+                    snafu::ErrorCompat::backtrace(self)
+                }
+            }
+        } else {
+            quote! {}
+        };
+
         let error_impl = quote! {
             impl#generics snafu::Error for #parameterized_struct_name
             where
@@ -1698,6 +1719,7 @@ impl StructInfo {
                 #description_fn
                 #cause_fn
                 #source_fn
+                #std_backtrace_fn
             }
         };
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,6 @@
 #![deny(missing_docs)]
 #![cfg_attr(not(any(feature = "std", test)), no_std)]
+#![cfg_attr(feature = "unstable-backtraces-impl-std", feature(backtrace))]
 
 //! # SNAFU
 //!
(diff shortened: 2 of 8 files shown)


snafu-derive/src/lib.rs
```diff
@@ -1296,7 +1297,11 @@ impl<'a> quote::ToTokens for ContextSelector<'a> {
         let backtrace_field = match *backtrace_field {
             Some(ref field) => {
                 let name = &field.name;
-                quote! { #name: std::default::Default::default(), }
+                if cfg!(feature = "std-backtraces") {
+                    quote! { #name: snafu::Backtrace::capture(), }
```
ccfb3ee765a58cae: Should we use force_capture or capture? Just using capture means the backtrace will only be created if RUST_BACKTRACE=1 at runtime. See #84

src/lib.rs
```diff
@@ -441,7 +445,7 @@ impl<T> OptionExt<T> for Option<T> {
 /// ```
 pub trait ErrorCompat {
```
ccfb3ee765a58cae: Should we even bother exposing an `ErrorCompat` when this feature is enabled?

Keeping it enabled seems like it would allow the smoothest ride to future rust.

compatibility-tests/std-backtrace/tests/backtrace.rs
```diff
@@ -0,0 +1,76 @@
+#![feature(backtrace)]
```
ccfb3ee765a58cae: These tests are just copied and pasted from `tests/`. Tests specific to snafu with `std::backtrace::Backtrace` support may be useful.

snafu-derive/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![allow(unknown_lints, bare_trait_objects)]
 #![recursion_limit = "128"] // https://github.com/rust-lang/rust/issues/62059
+#![cfg_attr(feature = "std-backtraces", feature(backtrace))]
```
shepmaster: Why is this needed? I would have thought it isn't, as we are only generating code that would use this feature, not using it ourselves.

src/lib.rs
shepmaster: I think so, yes. For as long as we want to support backtraces on stable Rust before the standard library stabilizes, we'll need to have something. If we were to remove it, then the compatibility shim wouldn't be providing much value to those users.

snafu-derive/src/lib.rs
shepmaster: Yeah, this is something I'd struggled a bit with when thinking through the problem. I think that using `capture` is the right place to start; we might eventually add something like `#[snafu(backtrace(force))]` to allow opting to the other function.

yaahc: I just left a comment on this in another repo, but you could have it interpret `Backtrace` as meaning `force_capture` and `Option<Backtrace>` as a regular `capture`

shepmaster:
I hope to get to this soon! Please see [my comment](https://github.com/shepmaster/snafu/pull/157#issuecomment-539511054) about my plans.

shepmaster:
Here's [my proposed combination of the two branches](https://github.com/shepmaster/snafu/compare/backtrace-revamp), plus some reorganizing, refactoring, and small feature additions.

TL;DR:

- `snafu::Backtrace` always exists (thanks to @lucab)
- The feature flag `backtraces` makes the `Backtrace` type live and implemented by our own type. 
- The feature flag `backtraces-impl-backtrace-crate`  makes the `Backtrace` type live and a re-export of `backtrace::Backtrace`.
- The feature flag `unstable-backtraces-impl-std`  makes the `Backtrace` type live and a re-export of `std::backtrace::Backtrace` (thanks to @ccfb3ee765a58cae).
- There's a new trait for generating backtraces to abstract across these three implementations and the trait allows `Option<Backtrace>`, if you want to prevent capturing the backtrace unless an environment variable is set (thanks to @yaahc).

How do people feel about this direction?

shepmaster:
Thanks!

PR #174, opened 2019-09-21
Add editorconfig file

ccfb3ee765a58cae:
From https://editorconfig.org

>EditorConfig helps maintain consistent coding styles for multiple developers working on the same project across various editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles. EditorConfig files are easily readable and they work nicely with version control systems.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false


.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
```
shepmaster: This is surprising; are there Markdown files with trailing whitespace? I think I'd rather fix those than codify it.

shepmaster:
I believe I've fixed the spurious build failure; please rebase to get a clean build.

.editorconfig
booty-bumping: Amazingly, trailing whitespace is significant in markdown. It indicates that the next line should start with `<br>` rather than continue the paragraph. If your editor keeps deleting it over and over again, it can become a pretty gigantic headache if the intention is to have a `<br>`

This is the same config that several projects including clippy use, so it can't be too crazy :)

ccfb3ee765a58cae: Amazingly, trailing whitespace is actually significant in markdown. It indicates that the next line should start with `<br>` rather than continue the current paragraph. If your editor keeps mysteriously deleting trailing whitespace, it can turn into a fairly big headache if you intend to have a `<br>`.

This is the same config used by clippy and a few other projects, so I don't think it's too crazy :)

ccfb3ee765a58cae: Blank lines are for starting new paragraphs. Actual line breaks leave no gap between lines.

Of course, there's a design argument that could be made that line breaks should be avoided at all costs, but sometimes legitimate uses for `<br>` do pop up. Not too often, though.

shepmaster: > my editor strips all trailing whitespace in all files, including Markdown. Never had a problem.

Ditto. However, mine has a bit of cleverness such that if I open a file with "offensive" whitespace choices, it leaves them and doesn't perform any extra cleanup. That will be useful here if we ever make use of it. 

However, I pretty aggressively remove double spaces after periods, so hopefully I'll remember that they may be important in the future...

> Of course, there's a design argument that could be made that line breaks should be avoided at all costs, but sometimes legitimate uses for `<br>` do pop up. Not too often, though.

Yeah, I am not a big fan of line breaks and prefer to use other elements and adjust the CSS as appropriate, but Markdown is interesting in that we rarely have control over the complete CSS anyway.

This seems reasonable for now, and maybe there's a better solution we can find if this bites us.

Issue #154, opened 2019-08-14
Decide on Rust 1.18 vs 1.31 requirement


shepmaster:
syn/quote/proc-macro2 just [updated to version 1.0](https://github.com/dtolnay/syn/releases/tag/1.0.0) and now require Rust 1.31+. Upgrading would break *our* version requirements. 

I chose the current 1.18 as basically "how early could we go without being terrible". My idea was to support as many Rust versions as possible in order to allow a broad range of projects to use SNAFU. 

It may not be the case that anyone *cares* about these older versions, though. Rust 1.31 is nearly 9 months old, and is 5 stable releases behind. I have a gut feeling that most people track stable or a few stable versions behind. 1.31 also has the notable benefit of being the release of Rust 2018 and NLL, things that people hotly anticipated and wanted.

As a user of SNAFU, I'd appreciate it if you'd vote on **one of** the following comments to provide feedback on what Rust versions we should support.

shepmaster:
Please add a 👍 to **this** comment if you'd like SNAFU to continue supporting Rust 1.18.

shepmaster:
Please add a 👍 to **this** comment if you'd be OK with SNAFU only supporting Rust 1.31+.

BurntSushi:
For crates I maintain that a lot of people depend on, I try to be fairly conservative, and my rough ballpark heuristic is "lag about a year behind latest stable." I think 9 months is just fine personally.

Dylan-DPC:
Faced with a similar situation earlier, I often solve it by asking the question "is there any language feature from those versions (1.18 to 1.30) that you depend on? if no, bump + major release

shepmaster:
@Dylan-DPC I've read your comment a few times over the last few hours, but I'm still confused. I'll try restating this specific problem with the hopes that I've communicated something poorly.

SNAFU has a dependency on syn/quote/proc-macro2 (which I'll shorten to just "syn" for the rest of this comment). Those dependencies have recently released version 1.0, and chosen to only guarantee compatibility with Rust 1.31 and *do not compile* on Rust 1.18. The previous versions had compatibility back to Rust 1.15, but SNAFU itself required Rust 1.18.

### SNAFU does nothing and stays on the current version of syn

Pro:
- supports Rust 1.18+
- more people/projects can use SNAFU

Con:
- annoyance to the people that will have to compile multiple versions of syn
- *maybe* cause issues with parsing future Rust syntax 

### SNAFU upgrades to syn 1.0

[...]

PR #177, opened 2019-09-24
Update to Rust 2018, syn/quote/proc-macro2 1.0

### LINKED ISSUE #164: Upgrade syn/quote/proc-macro2 to 1.0
Assuming #154 decides to bump SNAFU's Rust requirement to 1.31, we'll be able to upgrade our syn, quote, and proc-macro2 dependencies to 1.0, which also require Rust 1.31.

https://github.com/dtolnay/syn/releases/tag/1.0.0

Some potential benefits mentioned in #145: "making more use of the powers given by the new versions, maybe mostly in quote. For example all of the repetition things mentioned are issues we encountered."

Potentially related, also from #145: "I've had a mild sense that snafu-derive's lib.rs is simply too big at this point and could benefit from some reorganization. Perhaps reviewing the code during such reorganization would be a good time."

tjkirch:
Decision made in #154.
Fixes #164.

This takes advantage of all of the improvements in syn/quote/proc-macro2 1.0 that I could find in the code, and switches to some Rust 2018 idioms, but more could probably be done with more familiarity or research.

Certainly some of the changes in my commits are debatable, for example removing vs. replacing the `rust_1_30` feature.  We don't have anything conditional on version after this, so I opted for the simpler result of removing it.

Unit and compatibility tests pass.


Cargo.toml
```diff
@@ -50,17 +47,17 @@ rust_1_30 = ["snafu-derive/rust_1_30", "doc-comment"]
...
-  "compatibility-tests/v1_30",
+  "compatibility-tests/v1_31",
   "compatibility-tests/futures-0.1",
   "compatibility-tests/futures",
   "compatibility-tests/without-backtrace",
 ]
 
 [dependencies]
 snafu-derive = { path = "snafu-derive", version = "0.5.0" }
+doc-comment = "0.3.1"
```
shepmaster: Let's keep `default-features = false`. This helps minimize compile time where possible.

compatibility-tests/v1_31/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "v1_31"
+version = "0.1.0"
+authors = ["Jake Goulding <jake.goulding@gmail.com>"]
+
+[dependencies]
+snafu = { path = "../..", default-features = false, features = ["backtraces"] }
+
+# These versions are the last versions to support Rust 1.31
+# <none yet>
```
shepmaster: Heh, looks like we spoke too soon. Evidently the backtrace crate needs to be pinned to an earlier version here.

tjkirch:
Trying to figure out why the 1.31 test passes locally but fails in CI...

compatibility-tests/v1_31/Cargo.toml
tjkirch: Yep!  Sorry, forgot to clean my compatibility-test Cargo.lock files.  I wonder if I can automate that somehow...

I found it requires 0.3.35 at most.  backtrace-sys doesn't have to be pinned.  I just pushed that fix.

Cargo.toml
tjkirch: Sorry about that; just pushed the fix in the 1.30->1.31 commit.

PR #185, opened 2019-10-09
fix nit-picky doc typo

prasannavl:


src/guide/philosophy.md
```diff
@@ -6,7 +6,7 @@ effectively use SNAFU.
 
 ## Categorize underlying errors by their context
 
-It should be easy to bin one underlying error type (such as
```
shepmaster: This isn't actually a typo. I'm using "bin" as a verb to mean "put into a box" / "sort" / "categorize". Can you think of a less confusing way of stating the same concept?

prasannavl: Ah, yikes, my bad. What do you think about `wrap` instead of either of them?

prasannavl: or `map`, may be?

shepmaster: How do you feel about "compartmentalize" or "classify"? I asked around at a meetup I'm at and those were suggested. We discussed "wrap" and "map", but weren't excited about "wrap" because it doesn't give the same sense of assigning one thing to multiple possible categories. "map" has some existing meaning in Rust-world with all the methods named `map`.

prasannavl: Sounds good. PR updated.

Issue #110, opened 2019-07-08
Maybe have a Debug derive for Errors


CryZe:
What I don't particularly like is that both Debug and Display are used for errors in Rust. Most libraries and functions only focus on either the Debug or Display of the error. .unwrap() and fn main() -> Result<(), Error> tend to print the Debug implementation of the error. However snafu only focuses on providing Display implementations. So you end up deriving Debug for your errors which results in a struct / enum representation getting printed instead of the error messages on .unwrap() or when main fails. This is super unfortunate. I think it makes sense to have a derive for Debug that simply forwards to the carefully crafted error messages that we already have via Display.

shepmaster:
An interesting idea! This could look something like:

```rust
#[derive(Snafu, SnafuDebug)]
enum Error { /* … */ }

// Or maybe
#[derive(Snafu)]
#[snafu(Debug)]
enum Error { /* … */ }

// Or maybe
#[derive(Snafu(Debug))] // Is this even possible?
enum Error { /* … */ }
```

Ideally, switching between built-in `Debug` and SNAFU-powered `Debug` should be trivial for when developing and you need to debug.

The generated implementation should be trivial:

```rust
// Include initial generics and bounds…
// May need `Self: Display`
impl Debug for Error {
    fn fmt(&self, f) -> R {
        Display::fmt(self, f)
    }
}
```

One thought while typing the examples is that it feels like its about debugging SNAFU itself, not implementing `Debug`. Not sure how to work around that.

killercup:
Oh, great timing! I just made https://github.com/killercup/snafu-cli-debug because I wrote that code several times now. It's very much an 0.1, and could definitely use more tweaking. Also feel free to just adopt the code into the snafu derive.

shepmaster:
> I wrote that code several times now

Neat, thanks for sharing!

A quick skim shows that it has some additional features I hadn’t thought about:

- Printing a causal chain
- Printing the backtrace

These make sense, but may also want to be possible to opt-out. Something to consider when figuring out which syntax to use to enable this.

We may also want to steal the “causal chain iterator” idea from (I forget which library… error-chain?) so that we could just write something like:

```rust
for cause in CauseIter::new(self) {
    ::std::writeln!(f, "\tcause: {}", source)?;
}
```

I think that would make sense as a little tool to provide for people to use and allow us to backport [`iter_sources`](https://doc.rust-lang.org/std/error/trait.Error.html#method.iter_sources) if it ever stabilizes.

[...]

PR #186, opened 2019-10-09
add slightly improved main example

prasannavl:
A fairly common `print_error` implementation as a part of the main example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,21 @@ fn process_data() -> Result<()> {
 fn unpack_config(data: &str) -> &str {
     "/some/path/that/does/not/exist"
 }
+
+fn main() {
+    if let Err(e) = process_data() {
+        print_error(&e);
+    }
+}
+
+fn print_error(e: &dyn std::error::Error) {
+    eprintln!("error: {}", e);
+    let mut cause = e.source();
+    while let Some(e) = cause {
+        eprintln!("caused by: {}", e);
+        cause = e.source();
+    }
+}
 ```
 
 Please see [the documentation][Doc] and the [user's guide][Guide] for


README.md
```diff
@@ -35,6 +35,21 @@ fn process_data() -> Result<()> {
...
+}
+
+fn print_error(e: &dyn std::error::Error) {
+    eprintln!("error: {}", e);
+    let mut cause = e.source();
+    while let Some(e) = cause {
+        eprintln!("caused by: {}", e);
+        cause = e.source();
+    }
+}
```
shepmaster: This is a good start, but I'd rather avoid having people need to copy-paste this code in all of their projects. instead, see [this comment](https://github.com/shepmaster/snafu/issues/110#issuecomment-509245702) where we discuss adding an iterator that does the core part of this function. Then, this would look like

```rust
fn main() {
    if let Err(e) = process_data() {
        eprintln!("error: {}", e);
        for cause in whatever_iterator_is_called(e) {
            eprintln!("caused by: {}", cause);
        }
    }
}
```

prasannavl: If I remember correctly, there was a cargo package that kinda does this. Don't recall the specific name. It used the rust [Termination trait]( `https://doc.rust-lang.org/std/process/trait.Termination.html)` and wrapped other Error types with an iterator around it. 

Of course, I like the idea of snafu providing an iterator directly somehow, but just thought it was a nice way to kick start it off for quick programs.. seen quite a few times, where people didn't use it correctly for simple programs, and made me do a face palm :)

shepmaster: > there was a cargo package

There's an unstable feature from the standard library ([`iter_chain`](https://doc.rust-lang.org/nightly/std/error/trait.Error.html#method.iter_chain) / [`iter_sources`](https://doc.rust-lang.org/nightly/std/error/trait.Error.html#method.iter_sources)) that basically I'd like to copy into SNAFU for now, add to the `ErrorCompat` trait, then replace with the standard library's version when it stabilizes.

Would you be interested in adding that support to improve this documentation?

prasannavl: Sounds good. I'm not sure if I'll be able to make time this week though.. Just opening up an issue for that leaving this PR edited and currently blocked on it. Please feel free to close it if you feel it'd rather be tackled as one unit otherwise :)

PR #203, opened 2020-01-07
Make the formatter argument name unlikely to clash with user names

### LINKED ISSUE #202: Unhygienic Display impl
Given this code:
```rust
use snafu::Snafu;

#[derive(Debug, Snafu)]
pub enum Error {
    #[snafu(display("{}", f))]
    Whatever {
        f: i32,
    }
}
```
`cargo build` produces:
```
error[E0599]: no method named `write_fmt` found for type `&i32` in the current scope
```
which is caused by the name `f` conflicting with the name of the `&mut Formatter` in the `Display` impl.

shepmaster:
Fixes #202

### DIFF
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -1149,6 +1149,16 @@ fn private_visibility() -> UserInput {
     Box::new(quote! {})
 }
 
+struct StaticIdent(&'static str);
+
+impl quote::ToTokens for StaticIdent {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        proc_macro2::Ident::new(self.0, proc_macro2::Span::call_site()).to_tokens(tokens)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
> readability of the macro expanded code

But who will read the expanded code? Only developer of snafu, right? And even very rarely I suppose. I personally would also drop the UUID part (i.e. just `__snafu_display_f`), but I don't think readability of expanded code is too important, especially if it never ends up in the auto-generated documentation. But maybe I'm missing something?

dtolnay:
- Anyone using a debugger.
- Anyone trying to understand what SNAFU does, which is hard enough already. `cargo expand` is commonly used for this.

shepmaster:
Thanks all! I'll go with a middle ground of `__snafu_display_formatter`, which is:

1. Unlikely to clash.
1. Semi-self-documenting if someone does happen to run into it.

PR #205, opened 2020-01-14
feat: `allow(single_use_lifetimes)` where applicable

ErichDonGubler:
Users can run into an issue compiling code if their project uses `#[deny(single_use_lifetimes)]` for some reasonable error definitions. This attempts to alleviate that.

### DIFF
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -1564,6 +1564,7 @@ impl<'a> quote::ToTokens for DisplayImpl<'a> {
 
         stream.extend({
             quote! {
+                #[allow(single_use_lifetimes)]
                 impl<#(#original_generics),*> core::fmt::Display for #parameterized_enum_name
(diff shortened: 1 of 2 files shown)


shepmaster:
Thanks! I made some [small changes](https://github.com/shepmaster/snafu/compare/ee05a1c4026a074c9a96d7fc1c1e963329fa3c58..ecae46cd11400f67ee2a2992acdaf71aa5126090) and force-pushed.

I:

- added a test for the opaque error type
- removed unneeded `allow`s
- formatted the code

Can you check to make sure that this still works in your original case?

ErichDonGubler:
@shepmaster: Works well for me in my testing. Glad it wasn't hard to fix up -- I knew it'd need some tweaks. :)

shepmaster:
Thanks!

shepmaster:
Released as 0.6.2

Issue #197, opened 2019-11-16
Confusion in docs about wether if backtrace feature is enabled


scileo:
Hello,
There are two information about the default state of the `backtrace` feature.
 - [snafu::Backtrace page](https://docs.rs/snafu/0.6.0/snafu/struct.Backtrace.html) says it is enabled by default,
 - [snafu::guide::feature_flags page](https://docs.rs/snafu/0.6.0/snafu/guide/feature_flags/index.html#backtraces) says it is disabled by default.

As such, one of these two pages is wrong. I am no expert in how features are declared in `Cargo.toml`, but I think the default behaviour is backtrace disabled by default.

If this issue is accepted, then I'll be very happy to provide you a PR to fix the documentation.

shepmaster:
> snafu::Backtrace page says it is enabled by default,

It does not say that; it actually says (emphasis mine)

> Backtrace functionality is **currently** enabled. 

“Currently” does not mean “default“. Could you suggest another wording that would be more understandable?

scileo:
Wow sorry, I though the documentation system in cargo is static and is always the same, no matter the features selected. Generating custom documentation according to feature flags is a very nice idea!

This confused me because I was reading online documentation, so "currently" had no sense to me.

Unfortunately, my english level is not that high, so the best way to me to say it would be:

> This documentation was compiled with Backtrace functionality enabled.

This is a *way heavier* style, but that's the best idea I have. It would allow someone with low experience, reading online, not to make my mistake.

However, I don't think my suggestion worth the cost: as you pointed out, it was a misunderstanding from my part and not a confusion from the documentation.

shepmaster:
It's worth pointing out that [we enable non-default features on docs.rs](https://github.com/shepmaster/snafu/blob/0.6.3/Cargo.toml#L23-L25)

PR #208, opened 2020-02-02
Update Backtrace documentation

Ayplow:
The guide has the backtraces feature listed as disabled, and it seems the inline documentation fell out of date. Just it case, it'd be a good idea to review similar places in the documentation.

Either that, or find a way to verify it? Maybe a doctest could force the explanation to keep up to date.

### DIFF
--- a/src/backtrace_shim.rs
+++ b/src/backtrace_shim.rs
@@ -3,8 +3,8 @@ use std::{fmt, path};
 
 /// A backtrace starting from the beginning of the thread.
 ///
-/// Backtrace functionality is currently **enabled**. Please review
-/// [the feature flags](crate::guide::feature_flags) to disable it.
+/// Backtrace functionality is currently **diabled**. Please review
+/// [the feature flags](crate::guide::feature_flags) to enable it.
 #[derive(Debug)]
 pub struct Backtrace(backtrace::Backtrace);
 


shepmaster:
I think this is a misunderstanding. The module you edited is the backtrace shim, which provides a newtype wrapper around the backtrace crate. When that module is enabled (via `backtraces`), then backtrace collection *is* enabled.

https://github.com/shepmaster/snafu/blob/0e18dbd730fcb0b6c207e0ae29246e32387d362d/Cargo.toml#L27-L41

https://github.com/shepmaster/snafu/blob/0e18dbd730fcb0b6c207e0ae29246e32387d362d/src/lib.rs#L94-L105

Without that specified, the "inert" backtrace type is present:

https://github.com/shepmaster/snafu/blob/0e18dbd730fcb0b6c207e0ae29246e32387d362d/src/lib.rs#L81-L92

Ayplow:
Ah I see, then this might not be the best way to improve the documentation. I found the source file through <https://docs.rs/snafu/0.6.2/snafu/struct.Backtrace.html> which seems to use the shim for its' source. As that isn't the default feature set, it wasn't clear how I was meant to enable the struct.

shepmaster:
> I found the source file through docs.rs

Yes, on docs.rs we [enable the backtrace feature by default](https://github.com/shepmaster/snafu/blob/0e18dbd730fcb0b6c207e0ae29246e32387d362d/Cargo.toml#L23-L25) (among other things) so that the documentation that most users will browse shows more of the features available.

shepmaster:
See also #197. Obviously this is a problem as multiple people have been confused about it, but I'm not sure how to solve the problem.

The only thing that comes to mind copy-pasting a paragraph about how Cargo has features and how docs.rs has a different set of features from the default into the docs of every function that makes use of feature flags.

Thoughts?

Plecra:
The features only change the behavior of the `Backtrace` struct, right? I think just using the same documentation between the different implementations should be enough. It'd sidestep the problem of choosing the feature set to compile the docs with too.

Issue #199, opened 2019-11-29
Snafu for enums with variant types


Migi:
# Motivation

Currently, Snafu is designed to work with enums like
```rust
enum MyEnum {
    VariantA {
        val1: Foo,
        val2: Bar
    },
    VariantB {
        val1: Baz,
        val2: i32
    },
}
```
However, a somewhat common pattern in Rust (examples [1](https://github.com/rust-lang/rust/blob/ad808d95c4839caedc2be76d0ed059dc920ab4b6/src/librustc/hir/mod.rs#L2752), [2](https://www.reddit.com/r/rust/comments/2rdoxx/enum_variants_as_types/cnezmon/)) is to have enums like
```rust
struct VariantA {
    val1: Foo,
    val2: Bar
}
struct VariantB {
    val1: Baz,
    val2: i32
}
enum MyEnum {
    VariantA(VariantA),
    VariantB(VariantB)
}
```
While it is slightly more verbose, it has a few advantages:

1. Since enum variants are not types, it's not possible to accept a `MyEnum::VariantA` as a function parameter or to store it somewhere. You can only pass/store `MyEnum`. However, when you've made each variant its own type, `VariantA` *can* be passed around or stored.

    ***Note:** there is [an RFC](https://github.com/rust-lang/rfcs/pull/2593) which proposes to add variant types to Rust, which would remove this advantage. However, [this comment](https://github.com/rust-lang/rfcs/pull/2593#issuecomment-486997257) doesn't give the impression that this will be implemented soon.*

2. If multiple enums can have the same variant, you must repeat that variant (and all its fields) in each enum definition. With the second style, you just repeat the `VariantA(VariantA)` line.

3. When matching on this kind of enum, if a variant has a lot of fields, in the first style you have to match like
    ```rust
    match my_enum {
        VariantA{ val1, val2, val3, val4, val5, .. } => /* ... */
    }
    ```
    whereas if you use the second style, you can match like ([example](https://github.com/rust-lang/rust/blob/c4375c9dfdd7f31de909f6e9384bac1bf37b44da/src/librustc_typeck/collect.rs#L871-L897))
    ```rust
    match my_enum {
        VariantA(a) => /* use a.val1, a.val2, etc */
    }
    ```

[...]

Issue #188, opened 2019-10-22
Option to put generated context selectors into a submodule


LPGhatguy:
First, thanks for making SNAFU, I'm really enjoying its ergonomics!

I started a new project and tried to create a couple nested errors, but hit a name collision between the selectors that SNAFU was generating and one of my existing structs.

My inner module looks like this:

```rust
pub struct Config {
    // ...
}

#[derive(Debug, Snafu)]
pub enum ConfigError {
    // ... 
}

impl Config {
    pub fn load() -> Result<Self, ConfigError> {
        // ...
    }
}
```

My outer module imports a couple things from this struct and exports a new error type that wraps `ConfigError`:

```rust
use crate::{Config, ConfigError};

#[derive(Debug, Snafu)]
pub enum SyncError {
    Config { // oh no, a name collision!
        source: ConfigError,
    }
}

pub fn sync() -> Result<(), SyncError> {
    let config = Config::load().context(Config)?; // Collision here, too!

    // ...
}
```

The `Snafu` derive macro on `SyncError` tries to define a struct named `Config` as the context selector, which ends up colliding with the struct `Config` that I imported before!

One solution for this particular case I thought of was to put my SNAFU error into its own `mod` block to scope everything the macro defines. Combined with `#[snafu(visibility)]` and re-exporting the error type, this seems to work:

```rust
use crate::Config;

mod error {
    use crate::config::ConfigError;
    use snafu::Snafu;

    #[derive(Debug, Snafu)]
    #[snafu(visibility = "pub(super)")]
    pub enum SyncError {
        Config {
            source: ConfigError,
        },
    }
}

pub use error::SyncError;

pub fn sync() -> Result<(), SyncError> {
    let config = Config::load().context(error::Config)?;

    // ...
}
```

Another way we could solve this is by adding a configuration option like `#[snafu(mod = "error")]` to instruct SNAFU to wrap all of its context selectors into a module with the given name.

Does this feature make sense? Is there a better solution to this particular problem that I'm not seeing?

Thanks!

shepmaster:
Yes, I think such an option does make sense. I expect that most people who have run into the problem have done as you suggest: put it in a module "by hand" and perform the relevant exports.

I think that your general sketch of the resulting code also is the good default when this is opted-in-to: create a module (name chosen by user), put all the context selectors in that module, then mark them as `pub(super)`. The error itself stays where it is.

There will undoubtedly be some nuances that are run into during development, but this seems like it's in a good state to start work on.

[...]

PR #210, opened 2020-03-02
Allow the context selector struct name to be specified

Migi:
You could already specify that no context selector should be generated, using `context(false)`. Now you can specify the name of the context selector, using `context(MySelectorTypeName)`.

Helps with #188. I mean, it doesn't do exactly what asked there, which was to be able to put the generated context selectors into a submodule, but in the comments of that issue [it was suggested](https://github.com/shepmaster/snafu/issues/188#issuecomment-567288348)) that this could be an alternative solution for the same issues.

This is also the first step toward #199, where custom context selectors will be needed.


shepmaster:
> This is also the first step toward #199, where custom context selectors will be needed.

Can you explain a bit more about the general path you see? I don't follow how this plays into that bigger feature.

src/guide/attributes.md
```diff
@@ -46,6 +46,38 @@ fn main() {
 }
 ```
 
+## Specifying the generated context selector type's name
+
+Sometimes, you may already have a struct with the same name as one of
+the variants of your error enum. You might also have 2 error enums
+that share a variant. In those cases, there may be naming collisions
+with the context selector type(s) generated by Snafu, which by default
```
shepmaster, line 54: The proper (non-code) name of the library is "SNAFU".

src/guide/attributes.md
```diff
@@ -46,6 +46,38 @@ fn main() {
...
+# use snafu::{Snafu, ResultExt};
+#
+// some struct not related to the error
+struct Foo;
+
+#[derive(Debug, Snafu)]
+enum Error {
+    #[snafu(context(FooContext))]
+    Foo {
+        source: std::io::Error
```
shepmaster, line 72:
```suggestion
        source: std::io::Error,
```

tests/custom_selector_name.rs
```diff
@@ -0,0 +1,59 @@
...
+    },
+    #[snafu(
+        context(SaveConfigContext),
+        display = r#"("Could not open config file at {}", source)"#
+    )]
+    SaveConfig { source: io::Error },
+    #[snafu(
+        context(InvalidUserContext),
+        display = r#"("User ID {} is invalid", user_id)"#
+    )]
```
shepmaster, line 27: This style is fine, which is why it compiles, but I just find it ugly for whatever reason. FWIW, I tend to split it into two:

```rust
    #[snafu(context(InvalidUserContext))]
    #[snafu(display = r#"("User ID {} is invalid", user_id)"#)]
```

tests/custom_selector_name.rs
```diff
@@ -0,0 +1,59 @@
...
+        context(OpenConfigContext),
+        display = r#"("Could not open config file at {}: {}", filename.display(), source)"#
+    )]
+    OpenConfig {
+        filename: PathBuf,
+        source: io::Error,
+    },
+    #[snafu(
+        context(SaveConfigContext),
+        display = r#"("Could not open config file at {}", source)"#
```
shepmaster, line 21: Feel free to use the modern style in these tests:

```rust
        display("Could not open config file at {}", source)
```

tests/custom_selector_name.rs
```diff
@@ -0,0 +1,59 @@
...
+use snafu::{ResultExt, Snafu};
+use std::{
+    fs, io,
+    path::{Path, PathBuf},
+};
+
+#[derive(Debug, Snafu)]
+enum Error {
+    #[snafu(
+        context(OpenConfigContext),
```
shepmaster, line 12: Seeing it in use, I wonder if we want to give some more "namespacing". For example:

```rust
        context(name(OpenConfigContext)),
```

Or

```rust
        context(selector(OpenConfigContext)),
```

Or

```rust
        context(selector_name(OpenConfigContext)),
```

PR #250, opened 2020-09-13
Support struct errors

shepmaster:


shepmaster:
@tjkirch, @Enet4 — this is a big ole hunk of code that adds both a small and big feature to SNAFU. You should feel **no obligation** to review this, but if you *want* to, I'd love to hear your feedback! I'll probably merge and release this this weekend unless something is found.

Enet4 (commented):
I mostly skimmed through the guide and tests, things look good!

snafu-derive/src/lib.rs
```diff
@@ -1339,492 +1422,315 @@ impl<'a> quote::ToTokens for ContextSelectors<'a> {
     }
 }
 
-struct ContextSelector<'a>(&'a EnumInfo, &'a VariantInfo);
+struct ContextSelector<'a>(&'a EnumInfo, &'a FieldContainer);
```
tjkirch, line 1425: I don't see a problem here, but if you'd help me understand it a bit better, I'd appreciate it!

The part of SNAFU I have the most trouble understanding is the code generation, both because I haven't worked in that area of SNAFU and because I haven't done much of it in other crates.  So, I'm sorry if this is pretty basic.

What does this `ContextSelector` do?  I see that it's specific to enums, as is `ContextSelectors`, and I don't see something similar for `ToTokens` for structs.  Is this related to generating the selector structs for the enum variants?  I'm guessing so, because the `generate_snafu` implementation for enums relies heavily on `ContextSelectors`, whereas for structs it does a lot more work directly in the function using the fields and field containers, to build up a `ContextSelector` in the end.

compatibility-tests/v1_31/src/lib.rs
```diff
@@ -1 +1,97 @@
+#![cfg(test)]
+
 // There are not yet any feature differences that require testing for Rust 1.31 compatibility.
```
tjkirch, line 3: Out of curiosity, what are these new v1.31 "*_style" modules checking for?  It seems like a basic functionality test, but I'm not sure how it relates to 1.31 specifically.

Is this comment still accurate about having no feature differences to check?

src/guide/structs.md
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+}
+```
+
+## Differences from enum errors
+
+While each enum error variant creates a context selector that matches
+the variant's name, context selectors for structs remove the suffix
+`Error` from the name of the error, if present,  and add `Context`:
```
tjkirch, line 43: It did take me a bit to understand why this renaming happened, but I think I understand now.  Probably the most direct reason is that SNAFU couldn't generate selector structs with the same name in the same scope.  Do I have that right?  If so, I'm not sure it would work to add `Snafu` to a struct that's not suffixed with `Error`.

I think the approach you have makes sense.

Out of curiosity, are there any other approaches you considered?  One that jumped to my mind, though it might be silly, is to use namespacing.  Generated selectors could be placed in a new module, either with a name given as an attribute or a default.  I'm not sure it makes sense to have multiple implementations, though, and renaming is certainly simpler, even if it causes a question or two.

tjkirch (approved):
Thanks for inviting me!  I reviewed and everything seems to be up to your normal *very* high standards.  I don't see anything that makes me think it's not ready to ship.  I just have a few questions below for my own curiosity.

snafu-derive/src/lib.rs
shepmaster: > The part of SNAFU I have the most trouble understanding is the code generation, both because I haven't worked in that area of SNAFU and because I haven't done much of it in other crates. 

It's also one of the more gnarly parts of this code, IMO. I always have to spend a bit of time thinking about it.

> What does this `ContextSelector` do? I see that it's specific to enums, as is `ContextSelectors`, and I don't see something similar for `ToTokens` for structs. Is this related to generating the selector structs for the enum variants? I'm guessing so, because the `generate_snafu` implementation for enums relies heavily on `ContextSelectors`, whereas for structs it does a lot more work directly in the function using the fields and field containers, to build up a `ContextSelector` in the end.

The `quote!` macro can be used like `quote! { #something }`. Anything that implements [`ToTokens`](https://docs.rs/quote/1.0.7/quote/trait.ToTokens.html) can be used for interpolation. This `ContextSelector` basically just exists as a place to implement `ToTokens` on.

Previously, I had extracted this code in an attempt to help organize it. However, that organization wasn't quite what was needed to to support structs. I re-extracted it into the types in the `shared` module. The struct code uses those new types directly, and the enum code still goes through the old types. It seems likely that I'll try to unify those two code paths a bit more, but that's mostly refactoring and I didn't want to get stuck in a refactoring hole forever. :-)

compatibility-tests/v1_31/src/lib.rs
shepmaster: Yeah, still no feature differences. 

I realized that this test was checking that `snafu` and `snafu-derive` would compile in Rust 1.31, but _wasn't_ testing that the code **generated** by the macro would compile in Rust 1.31 because it wasn't generating any code. This just adds some basics.

src/guide/structs.md
shepmaster: > that SNAFU couldn't generate selector structs with the same name in the same scope. Do I have that right?

exactly!

> I'm not sure it would work to add `Snafu` to a struct that's not suffixed with `Error`.

I'm not totally following what you mean here.

> Generated selectors could be placed in a new module,

This is something that's been brought up before, but I don't immediately see how it can work because procedural macros only have access to the tokens they are on. For example:

```rust
use my::thing::Foo;

#[derive(Snafu)]
enum Error {
    Alpha { foo: Foo },
}
```

We only see `Foo`, not `my::thing::Foo`, so if we generated a submodule, `Foo` would no longer be in scope:

```
mod context {
    struct Alpha { foo: Foo };
}
```

Typing that out though, the selector is actually all generic types, and we could put the implementations outside of the module:

```
use my::thing::Foo;

#[derive(Snafu)]
enum Error {
    Alpha { foo: Foo },
}

mod context {
    struct Alpha<F> { foo: F };
}

impl<F: Into<Foo>> context::Foo {
    // ...
}
```

So it might indeed be possible.

---

I see a few possible futures:

1. Change enums to *also* add `Context` to the context selector. This makes the cases uniform, and *might* reduce one of the biggest points of confusion about using SNAFU — the usage of the same name to mean two different things.
1. Add the ability to control the selector name / how it's generated anyway. I might like to prefer `Ctx` for my own crates, instead of typing out the long form.
1. Add the ability to namespace the selectors, as you mentioned. We'd have to figure out how to support multiple errors creating modules in the same parent module, as otherwise the new modules would conflict.

compatibility-tests/compile-fail/tests/ui/structs/attribute-misuse.stderr
```diff
@@ -0,0 +1,35 @@
+error: expected `display`, `visibility`, `source`, `backtrace`, `context`, or `crate_root`
+ --> $DIR/attribute-misuse.rs:5:13
+  |
+5 |     #[snafu(unknown_attribute)]
+  |             ^^^^^^^^^^^^^^^^^
+
+error: `source` attribute is only valid on enum variant fields, not on a named struct
```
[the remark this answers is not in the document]
shepmaster: @tjkirch great catch! How does this read to you?

```diff
diff --git a/compatibility-tests/compile-fail/tests/ui/attribute-misuse.stderr b/compatibility-tests/compile-fail/tests/ui/attribute-misuse.stderr
index 64f0366..2e59a33 100644
--- a/compatibility-tests/compile-fail/tests/ui/attribute-misuse.stderr
+++ b/compatibility-tests/compile-fail/tests/ui/attribute-misuse.stderr
@@ -1,46 +1,46 @@
-error: `display` attribute is only valid on enum variants, not on an enum
+error: `display` attribute is only valid on enum variants or structs with named fields, not on an enum
  --> $DIR/attribute-misuse.rs:5:13
   |
 5 |     #[snafu(display("display should not work here"))]
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: `source(from)` attribute is only valid on enum variant fields, not on an enum
+error: `source(from)` attribute is only valid on enum variant or struct fields with a name, not on an enum
  --> $DIR/attribute-misuse.rs:6:13
   |
 6 |     #[snafu(source(from(XXXX, Box::new)))]
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: `source(bool)` attribute is only valid on enum variant fields, not on an enum
+error: `source(bool)` attribute is only valid on enum variant or struct fields with a name, not on an enum
  --> $DIR/attribute-misuse.rs:7:13
   |
 7 |     #[snafu(source(true))]
   |             ^^^^^^^^^^^^
 
-error: `backtrace` attribute is only valid on enum variant fields, not on an enum
+error: `backtrace` attribute is only valid on enum variant or struct fields with a name, not on an enum
  --> $DIR/attribute-misuse.rs:8:13
   |
 8 |     #[snafu(backtrace)]
   |             ^^^^^^^^^
 
-error: `context` attribute is only valid on enum variants, not on an enum
+error: `context` attribute is only valid on enum variants or structs with named fields, not on an enum
  --> $DIR/attribute-misuse.rs:9:13
   |
 9 |     #[snafu(context)]
   |             ^^^^^^^
 
-error: `source` attribute is only valid on enum variant fields, not on an enum variant
+error: `source` attribute is only valid on enum variant or struct fields with a name, not on an enum variant
   --> $DIR/attribute-misuse.rs:21:46
    |
 21 |         #[snafu(display("an error variant"), source(from(XXXX, Box::new)))]
    |                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: `source` attribute is only valid on enum variant fields, not on an enum variant
+error: `source` attribute is only valid on enum variant or struct fields with a name, not on an enum variant
   --> $DIR/attribute-misuse.rs:22:17
    |
 22 |         #[snafu(source)]
    |                 ^^^^^^
 
-error: `backtrace` attribute is only valid on enum variant fields, not on an enum variant
+error: `backtrace` attribute is only valid on enum variant or struct fields with a name, not on an enum variant
   --> $DIR/attribute-misuse.rs:23:17
    |
 23 |         #[snafu(backtrace)]
@@ -52,13 +52,13 @@ error: `crate_root` attribute is only valid on an enum or a struct, not on an en
 24 |         #[snafu(crate_root(XXXX))]
    |                 ^^^^^^^^^^^^^^^^
 
-error: `display` attribute is only valid on enum variants, not on a field
+error: `display` attribute is only valid on enum variants or structs with named fields, not on a field
   --> $DIR/attribute-misuse.rs:35:21
    |
 35 |             #[snafu(display("display should not work here"))]
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: `visibility` attribute is only valid on an enum and its variants, not on a field
+error: `visibility` attribute is only valid on an enum, enum variants, or a struct with named fields, not on a field
   --> $DIR/attribute-misuse.rs:36:21
    |
 36 |             #[snafu(visibility(pub))]
@@ -70,7 +70,7 @@ error: Incompatible attributes [`source(false)`, `source(from)`] specified on a
 38 |             #[snafu(source(from(XXXX, Box::new)))]
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: `context` attribute is only valid on enum variants, not on a field
+error: `context` attribute is only valid on enum variants or structs with named fields, not on a field
   --> $DIR/attribute-misuse.rs:39:21
    |
 39 |             #[snafu(context)]
@@ -82,31 +82,31 @@ error: `crate_root` attribute is only valid on an enum or a struct, not on a fie
 40 |             #[snafu(crate_root(XXXX))]
    |                     ^^^^^^^^^^^^^^^^
 
-error: `display` attribute is only valid on enum variants, not on a tuple struct
+error: `display` attribute is only valid on enum variants or structs with named fields, not on a tuple struct
   --> $DIR/attribute-misuse.rs:53:13
    |
 53 |     #[snafu(display("display should not work here"))]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: `visibility` attribute is only valid on an enum and its variants, not on a tuple struct
+error: `visibility` attribute is only valid on an enum, enum variants, or a struct with named fields, not on a tuple struct
   --> $DIR/attribute-misuse.rs:55:13
    |
 55 |     #[snafu(visibility(pub))]
    |             ^^^^^^^^^^^^^^^
 
-error: `source(bool)` attribute is only valid on enum variant fields, not on a tuple struct
+error: `source(bool)` attribute is only valid on enum variant or struct fields with a name, not on a tuple struct
   --> $DIR/attribute-misuse.rs:56:13
    |
 56 |     #[snafu(source(true))]
    |             ^^^^^^^^^^^^
 
-error: `backtrace` attribute is only valid on enum variant fields, not on a tuple struct
+error: `backtrace` attribute is only valid on enum variant or struct fields with a name, not on a tuple struct
   --> $DIR/attribute-misuse.rs:57:13
    |
 57 |     #[snafu(backtrace)]
    |             ^^^^^^^^^
 
-error: `context` attribute is only valid on enum variants, not on a tuple struct
+error: `context` attribute is only valid on enum variants or structs with named fields, not on a tuple struct
   --> $DIR/attribute-misuse.rs:58:13
    |
 58 |     #[snafu(context)]
diff --git a/compatibility-tests/compile-fail/tests/ui/structs/attribute-misuse.stderr b/compatibility-tests/compile-fail/tests/ui/structs/attribute-misuse.stderr
index 500e082..2f98906 100644
--- a/compatibility-tests/compile-fail/tests/ui/structs/attribute-misuse.stderr
+++ b/compatibility-tests/compile-fail/tests/ui/structs/attribute-misuse.stderr
@@ -4,31 +4,31 @@ error: expected `display`, `visibility`, `source`, `backtrace`, `context`, or `c
 5 |     #[snafu(unknown_attribute)]
   |             ^^^^^^^^^^^^^^^^^
 
-error: `source` attribute is only valid on enum variant fields, not on a named struct
+error: `source` attribute is only valid on enum variant or struct fields with a name, not on a named struct
   --> $DIR/attribute-misuse.rs:13:13
    |
 13 |     #[snafu(source(true))]
    |             ^^^^^^^^^^^^
 
-error: `backtrace` attribute is only valid on enum variant fields, not on a named struct
+error: `backtrace` attribute is only valid on enum variant or struct fields with a name, not on a named struct
   --> $DIR/attribute-misuse.rs:14:13
    |
 14 |     #[snafu(backtrace)]
    |             ^^^^^^^^^
 
-error: `display` attribute is only valid on enum variants, not on a field
+error: `display` attribute is only valid on enum variants or structs with named fields, not on a field
   --> $DIR/attribute-misuse.rs:23:17
    |
 23 |         #[snafu(display("display should not work here"))]
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: `visibility` attribute is only valid on an enum and its variants, not on a field
+error: `visibility` attribute is only valid on an enum, enum variants, or a struct with named fields, not on a field
   --> $DIR/attribute-misuse.rs:24:17
    |
 24 |         #[snafu(visibility(pub))]
    |                 ^^^^^^^^^^^^^^^
 
-error: `context` attribute is only valid on enum variants, not on a field
+error: `context` attribute is only valid on enum variants or structs with named fields, not on a field
   --> $DIR/attribute-misuse.rs:25:17
    |
 25 |         #[snafu(context)]
```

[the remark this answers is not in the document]
shepmaster: It's better, but has the problem that English isn't a programming language. I'd prefer to write it as

> error: `display` attribute is only valid on (enum variants or structs) with named fields, not on an enum

But that's not really how it works.

src/guide/structs.md
tjkirch: >> I'm not sure it would work to add Snafu to a struct that's not suffixed with Error.

> I'm not totally following what you mean here.

I meant something very close to the first point:

> SNAFU couldn't generate selector structs with the same name in the same scope

because if you don't have a suffix of "Error", SNAFU wouldn't do any renaming, and the context selector struct would have the same name as the original struct.  It works for enums (as you say, "the usage of the same name to mean two different things") because the names of enum variants and the names of structs live in different scopes.  With this PR, the generated selector structs and the original structs (with `derive(Snafu)`) live in the same scope, so they can't have the same name, which seems to imply to me that the input structs *must* be suffixed with "Error".  Am I making a logical misstep? :)

---

> might reduce one of the biggest points of confusion about using SNAFU — the usage of the same name to mean two different things.

I did have trouble understanding that at first, and I've explained it to others, so I see where you're coming from.  I think any of the three options would solve that confusion.

I think (2) is a generalized case of (1), since it could have a default of adding "Context".  It doesn't *seem* like it'd be a lot of extra work to add the configurability, since there are somewhat similar examples already, like source transformation.

For (3), I admit I hadn't thought about multiple errors in a module, since I haven't used that yet.  It does seem like you'd want the module to be named the same for all errors, e.g. `mod context`.  With that restriction, a solution isn't jumping to mind...

I would lean toward going with what you have for now, with structs being special, and trying to think of a solution for (3)'s issues.

I worry that (1) and (2) (with adding "Context" by default) would be too much naming overhead.  I suspect that enums will still be the most common usage of SNAFU, and referencing them with something like `error::FooContext` rather than `error::Foo` would get annoying, as would (with (2)) having to specify that your desired naming transformation is to do nothing.  (Also, there's the cost of migration for existing users, but there could be ways of mitigating that, so it's not my first consideration.)

I suppose there's also the option of doing (2) just for structs, so users can tailor the special case to their liking.

compatibility-tests/compile-fail/tests/ui/structs/attribute-misuse.stderr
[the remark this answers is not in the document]
tjkirch: I think the the change above reads well, thanks!

tjkirch (approved):
The messaging changes look good to me!

Issue #264, opened 2020-11-11
Missing field source


IsaccBarker:
Hello! I have been trying to get Snafu to work for close to 3 hours now, and this is my last hope (:
Anyways, I've created the error ProjectError:
```rust
#[derive(Debug, Snafu)]
pub enum ProjectError {
    #[snafu(display("Unable to read configuration from {}", path))]
    IOConfigError {
        #[snafu(source(from(io::Error, Box::new)))]
        source: Box<io::Error>,
        path: &'static str,
    },
}
```

And I'm trying to use it like this:
```rust
fn foo() -> Result<(), ProjectError> {
  let src = fs::read(META_FILE).context(IOConfigError { path: META_FILE })?;
}
```

When I try and compile, I get `missing field 'source' in initializer of 'project::project_error::ProjectError'`. I also got `the trait bound 'project::project_error::ProjectError: snafu::IntoError<_>' is not satisfied`, however, the trait is bound, correct?

Anyways, sorry for the simple question, and thanks in advance.

Also, how do I apply labels? Is this project set so that non-maintainers cannot apply them? Just wondering.

IsaccBarker:
Found a solution. Instead of using `IOConfigError` or `ProjectError::IOConfigError`, you have to use `project_error::IOConfigError` (i.e. not `project_error::ProjectError::IOConfigError`). I may soon submit a pull request for a documentation addition.

shepmaster:
This is the single most confusing part of SNAFU, so more documentation is welcome.

This code:

```rust
#[derive(Debug, Snafu)]
pub enum ProjectError {
    #[snafu(display("Unable to read configuration from {}", path))]
    IOConfigError {
        #[snafu(source(from(io::Error, Box::new)))]
        source: Box<io::Error>,
        path: &'static str,
    },
}
```

expands into (some pieces/details removed):

```
#[derive(Debug, Snafu)]
pub enum ProjectError {
    IOConfigError {
        source: Box<io::Error>,
        path: &'static str,
    },
}

struct IOConfigError {
    source: ...,
    path: ...
}
```

It is this second type that you use as the argument to `context`. More details about the expanstion are [in the guide](https://docs.rs/snafu/0.6.9/snafu/guide/the_macro/index.html).

shepmaster:
> ```rust
>         #[snafu(source(from(io::Error, Box::new)))]
>         source: Box<io::Error>,
> ```

[...]

Issue #212, opened 2020-03-11
Add a troubleshooting section to the guide with common errors


shepmaster:
A common error is attempting to use `MyErrorEnum::Variant` when the context selector `Variant` should be used instead. It's worth documenting this.

Are there any other common issues?

0xpr03:
Another problem I've got:
I was always using something like `Err(MyError::ResponseLimit { response: result })`. That works well untill you try to emit a leaf error including a backtrace. Then you need `ResponseLimit { response: result }`. But you can't do `return Err(ResponseLimit { response: result });`, because that's not a `MyError`, it also won't work with `return Err(MyError::ResponseLimit(ResponseLimit { response: result }))`.
The fix is to use `return ResponseLimit { response: result }.fail();` and it works magically.

redragonx:
Another problem I found while learning SNAFU is that https://github.com/shepmaster/snafu/blob/2c2e6ea5617be02c06e0a8144a5e391de8cd067e/src/guide/attributes.md#controlling-visibility was nowhere to be found in the main docs.

Everything clicked after I found this guide.

shepmaster:
Using `.into()` inside a context selector causes

```
cannot infer type for type parameter `__T0` declared on the struct ...
```

e.g.

```rust
Foo { message: "hello".into() }.fail()
```

PR #265, opened 2020-11-11
Added multi-file brief example

IsaccBarker:
Attempts #264 

Added a brief example to the documentation about using multiple files. I'm adding this because I don't want others to waste 4 hours of their time like I did mine. I feel like this speaks for itself (correct me if I'm wrong👍).


shepmaster:
I'm always hesitant to add more details to the top-level documentation. My worry is that if everyone puts the specific pieces that are important to them there, then soon it will contain everything and there's no more organization.

I honestly don't know what is best to help people, so I don't think what you've done is _wrong_, but I wanted to explain my hesitation about merging it as-is. Is there anywhere in the guide that you think would be reasonable? Do you think this could be related to #212?

IsaccBarker:
@shepmaster You're right about it being related to #212. I'll stop this pull-request and starting working on a section like that.

PR #271, opened 2020-12-03
Disambiguate references to the core library from the ensure macro

### LINKED ISSUE #270: Could not find `convert` in `core` when using `ensure!` with a conflicting module called `core` in scope
```
error[E0433]: failed to resolve: could not find `convert` in `core`
   --> meridian-graphics/src/shader/user_uniform.rs:940:17
    |
940 | /                 ensure!(is_valid_glsl_identifier(name), InvalidUniformBlockFieldIdentifier {
941 | |                     uniform_block_name,
942 | |                     field_name: name
943 | |                 });
    | |___________________^ could not find `convert` in `core`
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
Looks as if the issue is that it considers `core::convert` to be an ambiguous import; writing `use core::convert` in the same source file raises an error that it isn't sure which module to use. Replacing `core::convert` in the `ensure!` macro source with `::core::convert` appears to fix the issue, but I am not sure if that would be a breaking change for another reason.

shepmaster:
Fixes #270

### DIFF
--- a/compatibility-tests/v1_31/Cargo.toml
+++ b/compatibility-tests/v1_31/Cargo.toml
@@ -9,3 +9,4 @@ snafu = { path = "../..", default-features = false, features = ["backtraces"] }
 
 # These versions are the last versions to support Rust 1.31
 backtrace = "= 0.3.35"
+cc = "= 1.0.61"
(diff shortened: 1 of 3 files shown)


tpdickso:
Yep, it's exactly the fix I did to get it working on my local machine!

shepmaster:
Released in [0.6.10](https://crates.io/crates/snafu/0.6.10)

PR #286, opened 2021-05-03
Change the default context selector suffix to Snafu

shepmaster:


SamWilsn:
Hopefully dropping a comment here isn't too forward ^^;;

I routinely use a pattern like this with snafu:

### `src/error.rs`
```rust
pub(crate) mod parse_error {
    use snauf::Snafu;

    #[derive(Debug, Snafu)]
    #[snafu(visibility = "pub(crate)")]
    pub enum ParseError {
        SomeBadThing {
            source: SomeOtherError,
        }
    }
}

pub use self::parse_error::ParseError;
```

### `src/main.rs`

```rust
pub mod errors;

use self::errors::{parse_error, ParseError};

use snafu::ResultExt;

fn main() -> Result<(), ParseError> {
    do_something()
        .context(parse_error::SomeBadThing)?;

    Ok(())
}
```

tl;dr I think a `#[snafu(module = "parse_error")]` option might be better than a suffix (though I'm not super familiar with macro hygiene, so it might not be possible at all.)

shepmaster:
> Hopefully dropping a comment here isn't too forward ^^;;

Not at all!

> a `#[snafu(module = "parse_error")]` option

Presumably this would create the module and then place the error and selectors into it, similar to what is described in #188? Ultimately, I don't think that approach can work.

Note that your example uses `SomeOtherError`. That's never imported in the module, so it'd generate a compiler error (perhaps this is what you meant by hygiene?). Even if it was imported elsewhere, the macro only knows about the literal text that the `derive` is attached to, so we'd only see `SomeOtherError`, not the path it was imported from.

At best, you'd have to provide fully-qualified paths for many of the variant fields.

SamWilsn:
> Presumably this would create the module and then place the error and selectors into it, similar to what is described in #188?

Oh, yes. That's pretty much it exactly.

> Note that your example uses `SomeOtherError`. That's never imported in the module, so it'd generate a compiler error

Right. I would've needed to `use super::*;` for this to work.

> perhaps this is what you meant by hygiene?

I was speaking more about the restrictions on what identifiers escape a macro. I think there are some rules on it, and I'm absolutely not familiar with them.

> Even if it was imported elsewhere, the macro only knows about the literal text that the `derive` is attached to, so we'd only see `SomeOtherError`, not the path it was imported from.
>
> At best, you'd have to provide fully-qualified paths for many of the variant fields.

Would you be able to expand the derive macro to a module with either:
 1. `use super::{<types>};` where `<types>` is a comma delimited list of the error types from the enum, or
 2. `use super::*;`?

I doubt (1) would work with `::` in types (`crate::FooError` or `::other_crate::BarError`), but I _think_ (2) would usually work?

So:

```rust
use dependency::SomeOtherError;

#[derive(Debug, Snafu)]
#[snafu(module)] // Let's pretend that `module` does a camel to snake-case conversion, and `module(...)` lets you name it?
pub enum ParseError {
    SomeOther { source: SomeOtherError },
}
```

Would expand to something like:

```rust
use dependency::SomeOtherError;

#[derive(Debug)]
pub enum ParseError {
    // ...
}

pub(crate) mod parse_error {
    use super::*;

    pub(crate) SomeOther {
        // ...
    }
}
```

shepmaster:
> I _think_ \[`use super::*;`\] would usually work?

I can see that working in many cases, but it'd probably fall down for some other cases:

```rust
#[derive(Snafu,Debug)]
pub enum ParseError {
    Thing { source: super::Error },
}
```

> be better than a suffix

*I'm* not a fan of the module-focused view. My personal direction is that each module should have it's own error type and selectors. The selectors should only be used in that module.

That being said, it seems that both solutions can coexist. Would you be interested in (a) taking this discussion to #188 and/or (b) attempting to implement it?

PR #302, opened 2021-09-17
Add a prelude

shepmaster:


src/lib.rs
```diff
@@ -191,6 +191,23 @@
 //! }
 //! ```
 
+pub mod prelude {
+    //! Traits and macros used by most projects. Add `use
+    //! snafu::prelude::*` to your code to quickly get started with
+    //! SNAFU.
+
+    pub use crate::{ensure, whatever, OptionExt as _, ResultExt as _};
```
Enet4, line 199: It seems that `whatever` is only available with `std`.

```suggestion
    pub use crate::{ensure, OptionExt as _, ResultExt as _};
    #[cfg(feature = "std")]
    pub use crate::whatever;
```

PR #310, opened 2021-09-27
Add shorthand formatting for snafu(display)

### LINKED ISSUE #309: Support shorthand interpolation syntax for `#[snafu(display(...))]`
[`thiserror`](https://docs.rs/thiserror/1.0.29/thiserror/) supports a shorthand syntax for interpolating members into the display string.

It would be nice to have this feature in snafu as well, and would also reduce manual labor required for people migrating to snafu.

shepmaster:
Closes #309

### DIFF
--- a/tests/display-shorthand.rs
+++ b/tests/display-shorthand.rs
@@ -1,0 +1,87 @@
+use snafu::prelude::*;
+
+#[derive(Debug, Snafu)]
+enum Error {
+    #[snafu(display("ID {} denied", id))]
+    OnlyPositionalArguments { id: i32 },
+
+    #[snafu(display("Person {name} with ID {id} denied"))]
+    OnlyShorthandArguments { id: i32, name: &'static str },
+
+    #[snafu(display("Person {} with ID {id} denied", name))]
+    PositionalAndShorthandArguments { id: i32, name: &'static str },
+
+    #[snafu(display("Person {name:?} with ID {id:X} denied"))]
+    ShorthandArgumentsWithSpecs { id: i32, name: &'static str },
+
+    #[snafu(display("ID {id} denied"))]
+    FieldNotUsedAsNamedArgument { id: i32, name: &'static str },
+
+    #[snafu(display("ID {id} denied at {time}", time = 1 + 1))]
+    AdditionalNamedArguments { id: i32 },
+
+    #[snafu(display("Person {name} with ID {id} denied", id = 99))]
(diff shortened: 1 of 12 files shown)


tests/display-shorthand.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    #[snafu(display("ID {id} denied"))]
+    FieldNotUsedAsNamedArgument { id: i32, name: &'static str },
+
+    #[snafu(display("ID {id} denied at {time}", time = 1 + 1))]
+    AdditionalNamedArguments { id: i32 },
+
+    #[snafu(display("Person {name} with ID {id} denied", id = 99))]
+    RedefinedNamedArguments { id: i32, name: &'static str },
+}
```
Enet4, line 25: How hard would it be to support shorthand formatting inside documentation comments?

```suggestion

    /// Person {name} with ID {id} denied
    ArgumentsInComments { id: i32, name: &'static str },
}
```

shepmaster: > how hard would it be

I don't think it's a one-line change, but it shouldn't be *too* hard; something like moving the [parsing](https://github.com/shepmaster/snafu/blob/e0b6b71b564313678135ee964c7af28ee935c71f/snafu-derive/src/lib.rs#L659-L674) around and reusing the same format string inspection code to get the shorthand names.

More importantly, it seems... suspicious... that you'd want things like `{user}` in of your user-facing documentation. Why do you want to show that?

Enet4: > More importantly, it seems... suspicious... that you'd want things like `{user}` in your user-facing documentation. Why do you want to show that?

Well, for one, these fields are generally as visible as the enum variants themselves, so it does not seem far out to make the message template an integral part of the documentation itself. Even without this feature, it would not feel out of place to document a variant like this:

```rust
    /// The person called `name` with the ID `id` was denied
    ArgumentsInComments { id: i32, name: &'static str },
```

Another concern is that if someone was already using doc comments, adding or replacing variants with parameters which should be presented to the user still requires writing `#[snafu(display(...))]`, thus creating a mixture of doc comments and attributes in an error type ([example](https://github.com/Enet4/dicom-rs/blob/master/ul/src/association/client.rs#L19)).

shepmaster: Updated -- thoughts?

Enet4: Wonderful, looks great! Thank you!

PR #330, opened 2022-05-01
Add an `ensure_whatever` macro.

### LINKED ISSUE #328: how to combine ensure! and whatever!
[There's this example in the docs](https://docs.rs/snafu/latest/snafu/macro.whatever.html#without-an-underlying-error):

```rust
if moon_is_rising() {
    whatever!(
        "We are recalibrating the dynamos for account {}, sorry",
        account_id,
    );
}
```
What I would like to write is:

```rust
ensure!(
    !moon_is_rising,
    whatever!(
        "We are recalibrating the dynamos for account {}, sorry",
        account_id,
    )
)
```

This is currently not possible.
I want to avoid is to write an if statement and use `ensure!` consistently in my code.

shepmaster:
This is a combination of `ensure` and `whatever`.

Closes #328

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,7 +207,7 @@ pub mod prelude {
     }
 
     #[cfg(any(feature = "std", test))]
-    pub use crate::whatever;
+    pub use crate::{ensure_whatever, whatever};
 
     #[cfg(feature = "futures")]
     pub use crate::futures::{TryFutureExt as _, TryStreamExt as _};
@@ -333,6 +333,8 @@ pub use no_std_error::Error;
 /// Ensure a condition is true. If it is not, return from the function
 /// with an error.
 ///
+/// ## Examples
+///
 /// ```rust
 /// use snafu::prelude::*;
 ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -448,6 +459,45 @@ macro_rules! whatever {
...
+/// #[snafu(whatever, display("Error was: {message}"))]
+/// struct Error {
+///     message: String,
+/// }
+/// type Result<T, E = Error> = std::result::Result<T, E>;
+///
+/// fn get_bank_account_balance(account_id: &str) -> Result<u8> {
+/// # fn moon_is_rising() -> bool { false }
+///     ensure_whatever!(
+///         moon_is_rising(),
```
8573, line 483: Was it intended to negate the condition?  If I'm not mistaken, this changes it from `if moon_is_rising()` to (after macro expansion) `if !moon_is_rising()`.

shepmaster: Not intended, *per se*, but I’m ok with it. Since it’s just an example, whatever shows the feature easiest is a good choice, IMO. I don’t expect anyone will remember the example from one version to the next 😅

PR #331, opened 2022-05-01
Show how to make custom Whatever types Send / Sync

### LINKED ISSUE #322: Whatever is not async compatible
Following the instructions on the front page, I have

```rust
#[derive(Debug, Snafu)]
pub enum Error {
    #[snafu(whatever, display("{message}"))]
    Whatever {
        message: String,
        #[snafu(source(from(Box<dyn std::error::Error>, Some)))]
        source: Option<Box<dyn std::error::Error>>,
    },
```

This results in the error:

```
error: future cannot be sent between threads safely
  --> src/main.rs:45:61
   |
45 |       async fn get(&self, key: &str) -> Result<String, Error> {
   |  _____________________________________________________________^

...  |
   | |_____^ future created by async block is not `Send`
   |
   = help: the trait `Send` is not implemented for `(dyn snafu::Error + 'static)`
note: future is not `Send` as this value is used across an await
  --> src/main.rs:46:12
```

Removing the Whatever block from my error enum fixes the message.   I wasn't using whatever! in my async function but it was annoying stripping it from the rest of my code.

I'm a rust newbie, so problem is likely PBKAC, but in case it might affect others I raised an issue.   I've now purged whatever usage, so I'm fine with you closing the issue if that's appropriate.

shepmaster:
Closes #322

### DIFF
--- a/tests/stringly_typed.rs
+++ b/tests/stringly_typed.rs
@@ -246,3 +246,51 @@ mod struck {
         }
     }
 }
+
+mod send_and_sync {
+    use snafu::prelude::*;
+
+    #[derive(Debug, Snafu)]
+    enum Error {
+        #[snafu(whatever, display("{}", message))]
+        Whatever {
+            #[snafu(source(from(Box<dyn std::error::Error + Send + Sync>, Some)))]
+            source: Option<Box<dyn std::error::Error + Send + Sync>>,
+            message: String,
+        },
+    }
+
+    type Result<T, E = Error> = std::result::Result<T, E>;
+
+    #[test]
+    fn implements_error() {
+        fn check<T: std::error::Error>() {}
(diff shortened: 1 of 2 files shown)


tests/stringly_typed.rs
```diff
@@ -246,3 +246,51 @@ mod struck {
...
+    type Result<T, E = Error> = std::result::Result<T, E>;
+
+    #[test]
+    fn implements_error() {
+        fn check<T: std::error::Error>() {}
+        check::<Error>();
+    }
+
+    #[test]
+    fn impleemnts_send() {
```
Enet4, line 272:
```suggestion
    fn implements_send() {
```

shepmaster: 🤦

PR #334, opened 2022-05-24
rework Debug for Location struct

### LINKED ISSUE #333: Customize `Location` debug output to remove `_other`
It's noise and looking at the `Debug` format is pretty common for quick debugging.

hellow554:
`_other` is excluded in the debug output now.

closes #333


shepmaster (approved):
Thanks!

PR #343, opened 2022-07-25
Allow iter_chain to be used on no_std

OliverUv:
Hello!

This change will allow `no_std` users of snafu to make use of snafu's `ErrorCompat::iter_chain` function. Usage remains exactly the same as before.

I could not figure out an easy way to write an automatic test for this since the cargo test harness needs `std` (hence the `test` portion of these cfg attributes which already exist in the codebase `#[cfg(not(any(feature = "std", test)))]`)

 Here is the code I have used to test whether the feature works or not: https://github.com/OliverUv/snafu_api_experiment - it is a little larger than a minimal example since it also contains a small experiment to hack in dyn trait source errors. The relevant code is L114-157 of main.rs ( https://github.com/OliverUv/snafu_api_experiment/blob/7127727fb6a2efe6ecaa85c292309dfaff9e0aa9/src/main.rs#L114 )

### DIFF
--- a/src/error_chain.rs
+++ b/src/error_chain.rs
@@ -4,18 +4,18 @@
 ///
 /// Can be created via [`ErrorCompat::iter_chain`][crate::ErrorCompat::iter_chain].
 pub struct ChainCompat<'a> {
-    inner: Option<&'a dyn std::error::Error>,
+    inner: Option<&'a dyn crate::Error>,
 }
 
 impl<'a> ChainCompat<'a> {
     /// Creates a new error chain iterator.
-    pub fn new(error: &'a dyn std::error::Error) -> Self {
+    pub fn new(error: &'a dyn crate::Error) -> Self {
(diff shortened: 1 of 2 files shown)


shepmaster:
I've [force-pushed some changes](https://github.com/shepmaster/snafu/compare/a05a7f958a29cb47e248470c0acdc73157487f75..6030ec9646064661fa399f98df860ba129290ea5) as the crate already has an alias for `Error`:

https://github.com/shepmaster/snafu/blob/b552522ca92031a4450c914dce114013af383b7c/src/lib.rs#L324-L332

Can you double-check that this still works for your original case?

shepmaster:
I've rebased on `main` after taking care of some tangentially-related CI failures.

shepmaster:
I pulled down your repo and it appeared to work well with the changes I made, so I'll go ahead and merge. Thanks!

PR #344, opened 2022-09-14
Support the Provider API

shepmaster:


src/Snafu.md
```diff
@@ -370,6 +371,196 @@ enum Error {
...
+
+// We can get the source error and downcast it at once
+outer
+    .request_ref::<InnerError>()
+    .expect("Must have a source");
+
+// We can get the deepest backtrace
+outer
+    .request_ref::<InnerError>()
+    .expect("Must have a backtrace");
```
mikeando, line 449: I think this piece should use `.request_ref::<snafu::Bracktrace>()` rather than `InnerError`?

shepmaster: Thanks! Addressed in [this force-push](https://github.com/shepmaster/snafu/compare/2d827deff4fba3557916bcc20f9a295fef86a5b3..bde44dc2348c80f210312b0bac0ba3f16ae3c169)

shepmaster:
@Enet4 @hellow554 @ErichDonGubler @seritools @Stargateur @mikeando  👋 [^1]

This is a chunky new feature for SNAFU and I'd appreciate your feedback. I don't want to explain _too_ much upfront because I hope that [the documentation](https://deploy-preview-344--shepmaster-snafu.netlify.app/snafu/derive.snafu#providing-data-beyond-the-error-trait) explains it (and if not, I'd love some suggestions on how to improve that!).

The TL;DR is that the (unstable) provider API is the replacement for the (unstable and now removed) `Error::backtrace` method. It's also much more powerful and can unlock a number of possibilities. I think this implementation is _just_ about at a place where I'd feel comfortable releasing it as an unstable feature flag for SNAFU.

If you have the time, I'd love if you could read [the docs](https://deploy-preview-344--shepmaster-snafu.netlify.app/snafu/derive.snafu#providing-data-beyond-the-error-trait) and provide any feedback. If you have even more time, take a look at [the tests](https://github.com/shepmaster/snafu/pull/344/files#diff-3370d69ab3e18df20b3e2cd1cb0377bf7e13336210b314e8d694d6fe2db5ee09). If you have a total abundance of time to spend on me, feel free to look at the implementation as well (and remember that proc-macro code is nasty 😢). 

Thanks so much for all your time and help!

[^1]: I picked y'all because I know you and you've shown interest in helping SNAFU recently. If you'd rather I leave you alone, please let me know!

Stargateur:
Disclaimer, I almost never use `dyn` and don't like use it. I play dump on purpose with this review:

- I didn't understand that Provider API was from std at first read. Looked like a snafu only thing ("errors will implement the Provider API" => "errors will implement the Provider API from std")
- I think example with an enum would better show what this api is capable of
- What happen when many error provide the same type ? the doc have an example with "the deepest backtrace" but that mean it fondamental unclear at runtime what I will get, imagine many of my error have an id, I would get the deepest id ? how snafu determine what is the deepest ?
- Do I need nightly to use this feature ?
- [Automatically provided data](https://deploy-preview-344--shepmaster-snafu.netlify.app/snafu/derive.snafu#automatically-provided-data) and [Explicitly provided data](https://deploy-preview-344--shepmaster-snafu.netlify.app/snafu/derive.snafu#explicitly-provided-data) title doesn't match, shouldn't automatically be "implicit" ?

My concern as user:

- Do I pay for this feature if I don't use it ? How much cost this feature ?
- Is that a good idea ? Practical use ? Should have follow guideline about what an typical error should provide, is that even exist ?

shepmaster:
> I almost never use `dyn` and don't like use it.

Same here. The good news is that you only really need to pop over to the trait object to call the `request_*` methods. I'm also double checking to see if there's a way that it could be added directly to `std::error::Error`, instead of on the trait object. It _is_ possible to use a free function from the standard library and completely avoid the trait object:

```rust
let e = WebserverError;
std::any::request_value::<HttpCode>(&e);
```

> I think example with an enum would better show what this api is capable of

I certainly should provide at least _one_ example using an enum, but I'm not following why an enum example would show the capabilities **better** — can you expand on that point?

> What happen when many error provide the same type ? the doc have an example with "the deepest backtrace" but that mean it fondamental unclear at runtime what I will get, imagine many of my error have an id, I would get the deepest id ? how snafu determine what is the deepest ?

The `Provider` API works by types and you can indeed only return one piece of data for a type. When there are multiple providers for the same type, the one that is provided *first* will win. 

By default, SNAFU will gather the provided data from the `source` first, before providing any data from the current error. This can be overridden through the `priority` attribute, which will cause this error to provide that data first, before the source.

Any suggestions for how / where to work this information into the docs?

>  Do I need nightly to use this feature ?

Yes and no. You can add `#[snafu(provide)]` in any version of Rust that SNAFU supports. Only when the feature flag is enabled will that expand to useful code. This allows people to add `provide` to their code today and if something in the dependency tree enables the feature flag, errors just become better.

> [Automatically provided data](https://deploy-preview-344--shepmaster-snafu.netlify.app/snafu/derive.snafu#automatically-provided-data) and [Explicitly provided data](https://deploy-preview-344--shepmaster-snafu.netlify.app/snafu/derive.snafu#explicitly-provided-data) title doesn't match, shouldn't automatically be "implicit" ?

Yeah, probably. I also thought about "manually" as the counterpart to "automatically" — any opinions on that?

> Do I pay for this feature if I don't use it (with the feature activated) ? How much cost this feature?

At compile time:

With the feature **inactive**, there's the time for parsing and manipulating the data inside the procedural macro for handling `snafu(provide)`. With the feature **active**, there's also time taken to generate the `Error::provide` method call for each error and then further compilation of that.

At run time:

If someone calls `request_ref` / `request_value`, then it's more-or-less a O(N) walk through your error type and all the sources with each level doing a bunch of "are you looking for type `X`?`" checks.

> Is that a good idea ? Practical use ? Should have follow guideline about what an typical error should provide, is that even exist ?

No, I believe that this is so very new that there's not a community consensus. I expect that backtraces will be the first primary use case, but others that I've thought about are annotating an error with an HTTP error code (e.g. HTTP 404 vs HTTP 500) or with some externally-defined error code (e.g. `rustc`'s `E0000`).

I don't know exactly what SNAFU's role should be here — I show those examples in the docs, but I don't call attention to them as "best practice".

## My actions

- [ ] clarify Provider API is from std: "errors will implement the Provider API" => "errors will implement the Provider API from std"
- [ ] add one enum example
- [ ] clarify attribute can be used without nightly
- [ ] better align "automatically" vs "explciitly"

Enet4 (commented):
Thank you for the heads-up and for this great contribution! Supporting the provider API extensively like this is indeed a big deal.

I could only point out a few minor things about the documentation (see inline comments). Also took me a while to figure out how to run the tests, but this was probably just me not being familiar with how test suites are structured here.

src/Snafu.md
```diff
@@ -370,6 +371,196 @@ enum Error {
...
+
+#[derive(Debug, Snafu)]
+#[snafu(provide(Severity => Severity::Bad))]
+struct InnerError;
+
+#[derive(Debug, Snafu)]
+// We want to expose `name` as a `&str`, not `&String`
+#[snafu(provide(ref, str => name))]
+// We want to expose `enabled` as a `bool`, not an `Option<bool>`
+#[snafu(provide(opt, bool => *enabled))]
```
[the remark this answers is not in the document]
shepmaster: When the data is being provided, we get references to all of the fields. Since this example does _not_ use `ref` to denote it's returning a reference, it's returning a `bool` instead of `&bool`.

src/lib.rs
```diff
@@ -1407,6 +1411,7 @@ macro_rules! location {
 #[cfg(any(feature = "std", test))]
 pub struct Whatever {
     #[snafu(source(from(Box<dyn std::error::Error>, Some)))]
+    #[snafu(provide(false))]
```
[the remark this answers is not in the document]
shepmaster: Great observation! `snafu(source)` will control `Error::source`, so that will continue to work. The rest is more complicated...

Because we don't know a useful concrete type of the underlying source, we couldn't implement something like `request_ref::<TheInnerError>()`. We could implement `request_ref::<dyn Error>`. More importantly, we don't chain off to the provided values of the underlying error, which is something I should fix.

That will mean that I'll be adding a fourth option to the attribute, something like:

```rust
#[snafu(provide(opt, ref, chain, dyn std::error::Error => source.as_deref()))]
```

- `opt` —> it's optional
- `ref` —> it's a reference
- `chain` —> if this value is present, add all of its provided data too.

Thoughts on the name `chain`? I also was thinking maybe `delegate`, but am open to any suggestions.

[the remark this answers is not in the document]
shepmaster: This should now be working (still need to update the documentation to reflect it)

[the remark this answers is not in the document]
shepmaster: And minimal docs / examples added.

shepmaster:
@Stargateur I've attempted to address your points in [this force-push](https://github.com/shepmaster/snafu/compare/eb1b3fa0669c83975cc8d3ea2f5207c86e2b93f3..4546268906da1ac97884a86d8c8a2f200cfe2a49)

Stargateur:
The doc is way more clear, my only concern is the enum example doesn't explain what happen when `UserId` is not provided.

```rust
let e = NetworkUnreachable { }.build();
let e = &e as &dyn std::error::Error;
// would panic
let _= e.request_ref::<UserId>().unwrap();
```

But that may be too detailed maybe just a sentence like "`NetworkUnreachable` doesn't have `UserId` if you unwrap `e.request_ref::<UserId>()` you will panic.

shepmaster:
> the enum example doesn't explain what happen when `UserId` is not provided.

I've amended the example a bit with:

```rust
match e.request_ref::<UserId>() {
    // Present when ApiError::Login or ApiError::Logout
    Some(user_id) => {
        println!("{user_id:?} experienced an error");
    }
    // Absent when ApiError::NetworkUnreachable
    None => {
        println!("An error occurred for an unknown user");
    }
}
```

shepmaster:
For those watching along, I'm hoping to merge this sometime in the next few days, so if you have something you'd like to say, now would be a great time! ❤️

Enet4 (approved):
No more concerns on my side. Great work! :+1:

src/guide/feature_flags.md
```diff
@@ -77,6 +78,17 @@ is implemented.
...
+
+**default**: disabled
+
+When enabled, SNAFU-generated errors will implement the
+[`std::error::Error::provide`] method, allowing data to be retrieved
+using `request_ref` and `request_value` on a [`std::error::Error`]
+trait object reference.
+
+It is recommended that only applications make use of this feature.
+
```
mikeando, line 91: Is it worth linking back to the main section in `Snafu.md` from here?

mikeando (approved):
This looks really good to me.

Don't forget to remove the `reviewer notes` section in `Snafu.md`.

Stargateur (approved):

shepmaster:
[Force-pushed](https://github.com/shepmaster/snafu/compare/a9fd3d8ef948627b48596b0d8ddfd336d93427cc..498da7415cf0882bc376484a4c0d1de7c89ca018) to address comments from @seritools to split the combined example for `snafu(provide(...flags...` into separate sections and examples that can be linked to.

src/guide/feature_flags.md
shepmaster: Good idea!

shepmaster: Addressed in [this force-push](https://github.com/shepmaster/snafu/compare/498da7415cf0882bc376484a4c0d1de7c89ca018..9e9b01f75b049eeb878862e5d9e325f6c71071d8)

shepmaster:
Thanks all!

Issue #379, opened 2023-06-29
snafu-derive syn is still in v1.0 


Enet4:
A major version of syn (v2.0) was released a while ago. In my projects, all other syn dependents but snafu have done the jump. Would you be interested in a PR to migrate syn to version 2?

shepmaster:
Yeah, I've been eyeing that. The "problem" is that it also comes with a MSRV bump, which means SNAFU will have to have a MSRV bump and thus bump the minor version. Bumping the MSRV should also involve checking and cleaning things like the `rust_1_39` and `rust_1_46` features. It would also allow us to finally move to edition 2021. 

Rust 1.56 was 2021-10-21, so hopefully people will have been able to upgrade by now.

Have you tried bumping the syn version / any ideas what kind of changes are needed?

8573:
As a point of reference, Debian has Rust 1.63 in its stable release, 1.48 in "oldstable", and 1.41 in "oldoldstable" (long-term support out to [2024-06-30](https://wiki.debian.org/DebianReleases)).  Alpine Linux 3.15 and newer have Rust 1.56 and newer; Alpine 3.10, the oldest [not-completely-unsupported](https://alpinelinux.org/releases/) version covered by the [package catalog](https://pkgs.alpinelinux.org/packages?name=rust&branch=v3.10&repo=&arch=&maintainer=), has Rust 1.34. I'm not sure any other distro both packages Rust and is conservative in updating it similarly to Debian; Debian at least is the most famous. (I've checked Fedora, Ubuntu, OpenSUSE, Arch, Gentoo, NixOS, Guix.)

Enet4:
> Have you tried bumping the syn version / any ideas what kind of changes are needed?

Just gave it a try. The trickiest part was adjusting the `DocComment` definition to comply with the way that attributes are parsed in the new version (it's hard for me to explain why), the rest was renaming a type and changing a field access to a method call.

A good deal of the code changes would still be around requiring Rust 1.56, mostly to clean up conditional attributes and removing outdated content.
I pushed them all into [this branch](https://github.com/shepmaster/snafu/compare/main...Enet4:snafu:change/syn_2) which I hope do both of these well. Let me know if you would be willing to consider a PR with this.

shepmaster:
Sure, feel free to open that as a PR where we can iterate and comment. 

<details>
<summary>Comments I would have left and we can discuss on the PR</summary>

[...]

PR #381, opened 2023-07-02
Require minimum Rust 1.56

Enet4:
As discussed in #379, this PR does step 1: bump the hard minimum supported version of Rust to 1.56.

### Summary

- change rust-toolchain accordingly
- remove Cargo features 1_34 and 1_46
- simplify feature gated code
- update guide and documentation to no longer mention existing features
- mark Location as non_exhaustive, remove _other field
- change CI min version test task to use Rust 1.56
- remove v1.34 compatibility tests


Enet4:
Just replying to some of the [feedback](https://github.com/shepmaster/snafu/issues/379#issuecomment-1616704314) provided before the PR:

> We probably need to grep for (variations on) 1.34 / 1.39 / 1.46 to find places we talk about them in prose.

My greps yield no more results in need of updates. 

> Why did snafu-derive get a default = [] feature?

My mistake, I probably kept it in while I was experimenting with making the `rust_v*` features enabled by default. It's removed now.

> The yes_async::async_body module can probably be inlined.

IIUIC this will only be the case once we require a minimum version of Rust 1.61.

> fn track_caller() can probably be inlined as well

:+1: Done now.

---

By the way, should `rust_1_61` be enabled by default? I noticed that the compatibility page claims so, but that does not appear to be the case.

shepmaster:
Would you mind pulling [Update compile_error test snapshots](https://github.com/shepmaster/snafu/pull/381/commits/90733b173b3a8fb0e6fda22450c52c7e91b5c4dc) out into its own PR? It's not really about updating to 1.56, just that the current stable output has changed. (also I want it for #383 😇)

shepmaster:
> IIUIC this will only be the case once we require a minimum version of Rust 1.61.

Hmm, I thought it was because Rust 1.34 didn't know the `async` keyword (only 1.39+). That file *also* has a switch on 1.61, but that shouldn't be relevant for merging the file.

> By the way, should `rust_1_61` be enabled by default? I noticed that the compatibility page claims so, but that does not appear to be the case.

Hmm, there's two points here...

1. The 0.7 docs are probably just wrong. They also are missing the 1.39 feature flag from that table. 
2. Since we are upping the MSRV and releasing a new semver-incompatible version, we will probably want to change the default feature-flagged version to be whatever is stable when we release 0.8 (likely 1.70 /1.71 which will effectively mean 1.61 as there are no other flags since then).

I'm generally planning on releasing one more 0.7 so I can fix those docs (and maybe cause some merge conflicts 😬).

Enet4:
> Would you mind pulling [Update compile_error test snapshots](https://github.com/shepmaster/snafu/pull/381/commits/90733b173b3a8fb0e6fda22450c52c7e91b5c4dc) out into its own PR? It's not really about updating to 1.56, just that the current stable output has changed. (also I want it for #383 😇)

Makes sense indeed. I'll get to that once I'm able.

> Maybe also [rustfmt src/lib.rs](https://github.com/shepmaster/snafu/pull/381/commits/e2f173579ca38304282853ef67b33f0acfe53c8d) if that's more drift from stable.

That was caused by the removal of the non-exhaustive lock field `_other`. I might just fuse the commit with a fixup.

> Hmm, I thought it was because Rust 1.34 didn't know the `async` keyword (only 1.39+). That file _also_ has a switch on 1.61, but that shouldn't be relevant for merging the file.

I guess it's easy to try if the code builds on Rust 1.56 anyway, and make changes if so. I'll look into it.

> Hmm, there's two points here...
> 
> 1. The 0.7 docs are probably just wrong. They also are missing the 1.39 feature flag from that table.
> 2. Since we are upping the MSRV and releasing a new semver-incompatible version, we will probably want to change the default feature-flagged version to be whatever is stable when we release 0.8 (likely 1.70 /1.71 which will effectively mean 1.61 as there are no other flags since then).

OK, in that case I will not touch the documentation (which already claims that `rust_v1_61` would be enabled).

> I'm generally planning on releasing one more 0.7 so I can fix those docs (and maybe cause some merge conflicts 😬).

_\*git rebase intensifies\*_

shepmaster:
Looking good. I made one small tweak to wording and force pushed (also rebased for good measure).

shepmaster (approved):

PR #383, opened 2023-07-05
Extract an iterator that cleans redundant nested error text

shepmaster:
This allows the cleaning logic to be reused in other contexts, such as an HTML or JSON error report.

### DIFF
--- a/src/report.rs
+++ b/src/report.rs
@@ -272,33 +272,23 @@ impl<'a> ReportFormatter<'a> {
     fn cleaned_error_trace(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
         const NOTE: char = '*';
 
-        let mut original_messages = ChainCompat::new(self.0).map(ToString::to_string);
-        let mut prev = original_messages.next();
-
-        let mut cleaned_messages = vec![];
         let mut any_cleaned = false;
         let mut any_removed = false;
-        for msg in original_messages {
-            if let Some(mut prev) = prev {
-                let cleaned = prev.trim_end_matches(&msg).trim_end().trim_end_matches(':');
-                if cleaned.is_empty() {
+        let cleaned_messages: Vec<_> = CleanedErrorText::new(self.0)
+            .flat_map(|(_, mut msg, cleaned)| {
+                if msg.is_empty() {
                     any_removed = true;
-                    // Do not add this to the output list
-                } else if cleaned != prev {
-                    any_cleaned = true;
-                    let cleaned_len = cleaned.len();
-                    prev.truncate(cleaned_len);
-                    prev.push(' ');
-                    prev.push(NOTE);
-                    cleaned_messages.push(prev);
+                    None
                 } else {
-                    cleaned_messages.push(prev);
+                    if cleaned {
+                        any_cleaned = true;
+                        msg.push(' ');
+                        msg.push(NOTE);
+                    }
+                    Some(msg)
                 }
-            }
-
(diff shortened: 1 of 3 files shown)


src/report.rs
```diff
@@ -357,6 +347,94 @@ fn trace_cleaning_enabled() -> bool {
     !DISABLED.get(|| env::var_os(SNAFU_RAW_ERROR_MESSAGES).map_or(false, |v| v == "1"))
 }
 
+/// An iterator over an Error and its sources that removes duplicated
+/// text from the error display strings.
+///
+/// It's common for errors with a `source` to have a `Display`
+/// implementation that includes thier source text as well:
```
Enet4, line 354:
```suggestion
/// implementation that includes their source text as well:
```

tests/report.rs
```diff
@@ -162,3 +162,59 @@ struct TestFunctionError;
 fn procedural_macro_works_with_test_functions() -> Result<(), TestFunctionError> {
     Ok(())
 }
+
+#[track_caller]
+fn assert_cleaning_step(iter: &mut CleanedErrorText, text: &str, removed_text: &str) {
```
Enet4, line 167: Do we have test coverage for the case where a message contains `:`, but the subsequent part of the string represents an unreachable cause not exposed via `source`? No clean-up should happen in these cases.

shepmaster: Do you mean something like a leaf message saying “bad thing happened: restart your server”?

Enet4: Yes, or the case where the second part is indeed part of a sub-cause error message but the corresponding cause is deliberately not part of the error source chain.

shepmaster: I'm not sure I follow that second case. Do you mean something like this pseudocode?

```rust
#[snafu(display("Outer is better than Inner"))] // doesn't use `{source}` but just happens to end with the exact same text
struct Outer { source: Inner }

#[snafu(display("Inner"))]
struct Inner;
```

If so, I don't think there's any way to avoid that.

Enet4: I was thinking about the code below. I suspect the confusion is because one is unlikely to do this in snafu, but if for any reason we have a source error which we do not want to expose as an error, the error text cleaning routine should not clean the message.

```rust
#[derive(Debug, Snafu)]
#[snafu(display("Epic fail: {}", hidden_source))]
struct Outer {
   hidden_source: Inner,
}

#[snafu(display("I AM ERROR"))]
struct Inner;
```

The display message of an `Outer` should be "Epic fail: I AM ERROR", per the recommendations of the error handling WG.

---

The situation you just mentioned is a curious one, though! Could we have problems with the outer error's message ending with the exact message as its source-error message?

shepmaster: > we do not want to expose as an error

In that case, it would not be cleaned — by construction. This walks the `Error::source` chain to find the duplicate text. It'e effectively trying to retroactively apply the error handling WG's suggestion of (`Error::source` XOR "include it in `Display`) for errors that do both.

Calling `Error::source` on `Outer` will return `None`, so we will never get the `Display` string from `Inner`, so we will not use it to clean `Outer`'s text.

---

> Could we have problems with the outer error's message ending with the exact message as its source-error message?

It's always possible. That's why `snafu::Report` offers [a way to disable the cleaning](https://github.com/shepmaster/snafu/blob/d993c94e999f7c1201b11e750b803a30b7588cc3/src/report.rs#L337-L341). I do think that it is pretty unlikely, however.

PR #396, opened 2023-08-20
Stop automatically treating location as implicit

### LINKED ISSUE #339: Do not automatically treat `location` as implicit data
The error message generated is poor ("struct \`DemoSnafu\` has no field named \`location\`" / "the trait bound \`WhatShouldHaveBeenLocation: GenerateImplicitData\` is not satisfied") and it's unlikely to be a deeply used feature.

shepmaster:
Closes #339

### DIFF
--- a/compatibility-tests/compile-fail/tests/ui/attribute-misuse-opt-out-wrong-field.stderr
+++ b/compatibility-tests/compile-fail/tests/ui/attribute-misuse-opt-out-wrong-field.stderr
@@ -10,7 +10,7 @@ error: `backtrace(false)` attribute is only valid on a field named "backtrace", 
 19 |             #[snafu(backtrace(false))]
    |                     ^^^^^^^^^^^^^^^^
 
-error: `implicit(false)` attribute is only valid on a field named "location", not on other fields
+error: `implicit(false)` attribute has no effect
   --> tests/ui/attribute-misuse-opt-out-wrong-field.rs:31:21
    |
 31 |             #[snafu(implicit(false))]
--- a/compatibility-tests/futures/src/location.rs
+++ b/compatibility-tests/futures/src/location.rs
@@ -3,19 +3,20 @@ use snafu::{location, prelude::*, Location};
 
 #[derive(Debug, Copy, Clone, Snafu)]
 struct InnerError {
+    #[snafu(implicit)]
     location: Location,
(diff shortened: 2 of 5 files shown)


Enet4 (approved):
It may feel a bit odd that we need to explicitly mark it as `implicit`, but once it is made clear that the _implicit_ has to do with the properties being implicitly provided upon error generation, that confusion goes away.

As (AIUI) we cannot reliably identify whether the field `location` is indeed of type `snafu::Location`, I agree that we are better off reducing the chances of false positives.

shepmaster:
> cannot reliably identify whether the field `location` is indeed of type

That's correct. Proc macros only have knowledge of the literal source code, not any types or anything fancy.

mhnap:
> Proc macros only have knowledge of the literal source code, not any types or anything fancy.

Seems `thiserror` somehow handles this correctly, as the backtrace field can be named differently than `backtrace` but if it has the `std::backtrace::Backtrace` type it still will be recognized as backtrace (`provide` method will be generated).

Firstly, I wanted to propose making `Location` implicit by default, but after I found this issue :)

shepmaster:
>> Proc macros only have knowledge of the literal source code, not any types or anything fancy.
>
> Seems thiserror somehow handles this correctly

Can you please provide a code example demonstrating what you mean? I stated that the literal source code is all that the macro gets, but it seems like you are saying that is not true. If so, you should be able to write code where the error type has no literal words like "backtrace" and the automatic backtrace behavior still works:

```rust
use std::backtrace::Backtrace as Wubble;

#[derive(Error, Debug)]
pub struct MyError {
    msg: String,
    wubble: Wubble,
}
```

If you provide that, then I can go digging into thiserror to see how it works.

mhnap:
Yes, you're right. It will work if the field name differs from `backtrace` but the field type still needs to contain `Backtrace`. So, it indeed will not work for your example above.
It works for:
```rust
    #[derive(Error, Debug)]
    pub struct MyError {
        msg: String,
        wubble: std::backtrace::Backtrace,
    }
```
or:
```rust
    use std::backtrace::Backtrace as Backtrace;

    #[derive(Error, Debug)]
    pub struct MyError {
        msg: String,
        wubble: Backtrace,
    }
```
Sorry for misleading.

PR #397, opened 2023-08-20
ChainCompat now returns &'a (dyn Error + 'b) to allow downcasting

### LINKED ISSUE #370: CompatChain returns `&'a dyn Error + 'a` which cannot be downcast
It should be `&'a dyn Error + 'static`. Doing this means that the public constructor needs to be modified, making it technically a breaking change.

shepmaster:
Previously it unconditionally returned &'a (dyn Error + 'a) which was never downcastable. The additional lifetime can be `'static`, resulting in `&'a (dyn Error + 'static)`, which may be downcast.

Closes #370


Enet4 (approved):
Looks OK to me.

PR #405, opened 2023-09-08
Fix doc: codegen

akiradeveloper:
Fields other than source should be given generic types.

I found this in reading the doc and felt very confusing because this is directly contradict to this description 

> Each remaining field’s type has been replaced with a generic type.


shepmaster (approved):
Heh, that looks like I just straight-up stopped before getting to `ConfigValidationFailedSnafu`. Thanks!

PR #409, opened 2023-09-24
Implement `snafu(transparent)`

### LINKED ISSUE #406: Consider adding #[snafu(transparent)]
While working on the playground recently, I found myself wanting a "transparent" error. Someone has mentioned this to me before, but we don't seem to have an issue for it. 

The TL;DR of my case:

- 2 data types that are parsed from strings (`Edition`, `Mode`)
- 2 HTTP requests that use the data types (`Run`, `Build`)

In the original playground implementation, the two data types shared one error type (e.g. `enum Error { BadEdition, BadMode }`) but that caused some annoying circular type problems and runs contrary to the normal Rust practice of one error per parse (e.g. `ParseBoolError`). Moving them to separate types leads to a catch — what do you put as the `Display` implementation on the outer error?

Concretely, code like

```rust
if let Err(e) = parse_run("moo", "debug") {
    println!("Couldn't parse `Run`: {e}");
}
```

Would print out:

```
Couldn't parse `Run`: Edition: 'moo' is not a valid edition
```

The `Edition: ` is the auto-generated error message, but there's not really any better text to put here. Ideally I'd just completely *remove* that middle text and end up with ``Couldn't parse `Run`: 'moo' is not a valid edition``.

Since we would want to use the source's `Display`, we probably shouldn't return the source as an error itself, which is why this sounds like "transparent" to me.

### Rough suggested implementation

Adding `#[snafu(transparent)]` to an enum variant will delegate the implementation of `Display` and `Error::source` to the `source` field.

- This attribute should be an error in other places
- Should be an error to combine it with `display`

### Open questions

- This would often be paired with `context(false)` — would it **always**? Should `transparent` imply `context(false)`?

<details>
<summary>Complete code example</summary>

```rust
use snafu::prelude::*;
use std::str::FromStr;

// Data types

enum Edition {
    Rust2015,
    Rust2018,
}

impl FromStr for Edition {
    type Err = ParseEditionError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        Ok(match s {
            "2015" => Edition::Rust2015,
            "2018" => Edition::Rust2018,
            value => return ParseEditionSnafu { value }.fail(),
        })
    }
}

#[derive(Debug, Snafu)]
#[snafu(display("'{value}' is not a valid edition"))]
struct ParseEditionError {
    value: String,
}

enum Mode {
    Debug,
    Release,
}

impl FromStr for Mode {
    type Err = ParseModeError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        Ok(match s {
            "debug" => Mode::Debug,
            "release" => Mode::Release,
            value => return ParseModeSnafu { value }.fail(),
        })
    }
}

#[derive(Debug, Snafu)]
#[snafu(display("'{value}' is not a valid mode"))]
struct ParseModeError {
    value: String,
}

// Requests

#[allow(dead_code)]
struct Run {
    edition: Edition,
    mode: Mode,
}

fn parse_run(edition: &str, mode: &str) -> Result<Run, ParseRunError> {
    Ok(Run {
        edition: edition.parse()?,
        mode: mode.parse()?,
    })
}

#[derive(Debug, Snafu)]
enum ParseRunError {
    #[snafu(context(false))]
    Edition { source: ParseEditionError },

    #[snafu(context(false))]
    Mode { source: ParseModeError },
}

#[allow(dead_code)]
struct Build {
    edition: Edition,
    mode: Mode,
}

fn parse_build(edition: &str, mode: &str) -> Result<Build, ParseBuildError> {
    Ok(Build {
        edition: edition.parse()?,
        mode: mode.parse()?,
    })
}

#[derive(Debug, Snafu)]
enum ParseBuildError {
    #[snafu(context(false))]
    Edition { source: ParseEditionError },

    #[snafu(context(false))]
    Mode { source: ParseModeError },
}

// Usage

fn main() {
    if let Err(e) = parse_run("moo", "debug") {
        println!("Couldn't parse `Run`: {e}");
    }

    if let Err(e) = parse_build("2018", "cow") {
        println!("Couldn't parse `Build`: {e}");
    }
}
```

</details>

seritools:
Fixes #406


snafu-derive/src/shared.rs
```diff
@@ -693,6 +712,10 @@ pub mod error {
                         quote! {
                             #field_name.as_ref().map(|e| e.as_error_source())
                         }
+                    } else if *is_transparent {
+                        quote! {
+                            #field_name.as_error_source().source()
```
seritools, line 717: not sure if this needs a manual `::core::error::Error::source`

src/Snafu.md
```diff
@@ -189,6 +190,19 @@ fn do_something_unique() -> Result<i32, VeryUniqueError> {
 }
 ```
 
+### Delegate to the underlying error
+
+Composing many different error types is idiomatic, can however lead to unwanted nesting, both in the
+string representation and in the source chain. To avoid this, you can use `#[snafu(transparent)]` to
+delegate the `Display` and `Error::source` implementations to the underlying error, effectively
+removing the wrapping error struct or enum variant from the logical error chain.
```
shepmaster: First sentence is rough after the comma.

Riffing on ideas...

> Composing different error types is idiomatic, but can lead to unwanted nesting, both in the string representation and in the source chain. 

I think we should circle back here after we have the example as well, which might help solidify more specific cases where this is useful.

tests/transparent.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[derive(Debug, Snafu)]
+enum AlphaError {
+    AlphaDummy,
+}
+
+#[derive(Debug, Snafu)]
+enum BetaError {
+    BetaDummy { source: AlphaError },
+}
```
shepmaster, line 12: Feel free to make these struct errors instead of enums if there's no reason to have variants.

tests/transparent.rs
```diff
@@ -0,0 +1,97 @@
...
+
+fn check<T: std::error::Error>() {}
+
+#[test]
+fn implements_error() {
+    check::<ErrorTest>();
+
+    let error = example().unwrap_err();
+    // skpping `ErrorTest`, directly going yo BetaError's source
```
shepmaster, line 45: yo

tests/transparent.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[test]
+fn implements_display() {
+    let error = example().unwrap_err();
+    // no `Beta: ` prefix
+    assert_eq!(error.to_string(), "BetaDummy: AlphaDummy");
+}
+
+mod with_backtraces {
```
shepmaster, line 56: Did this module (and the next) come about because of some specific issue during implementation, or just as insurance?

snafu-derive/src/lib.rs
```diff
@@ -638,6 +640,15 @@ const ATTR_CRATE_ROOT: OnlyValidOn = OnlyValidOn {
     valid_on: "an enum or a struct",
 };
 
+const ATTR_TRANSPARENT: OnlyValidOn = OnlyValidOn {
```
shepmaster, line 643: We should add compile-fail tests for the "special" new errors. The transparent + context error cases especially, as that's kind of a new pattern and not as easy as e.g. one more usage of `AtMostOne`.

(`cd compatibility-tests/compile-fail && cargo test`).

snafu-derive/src/shared.rs
```diff
@@ -528,6 +528,20 @@ pub mod display {
 
             let source_field = selector_kind.source_field();
 
+            if field_container.is_transparent {
+                // transparent errors always have a source field
```
shepmaster, line 532: Should we have a test to exercise what happens if we don't have a source field?

tests/transparent.rs
```diff
@@ -0,0 +1,97 @@
...
+    use super::*;
+    use snafu::Backtrace;
+
+    #[derive(Debug, Snafu)]
+    enum Error {
```
shepmaster, line 61: Changing this one to a struct will allow us to double-dip and have a test that `transparent` works on structs.

shepmaster (changes requested):
Looking good, minor nits.

> that `transparent` works on structs

Which reminds me... how different is this implementation from what we do for an opaque error? Should there be some sort of unification?

src/Snafu.md
```diff
@@ -37,6 +37,7 @@ it is valid. Detailed information on each attribute is below.
 | `context(false)`                | Skips creation of the context selector, implements `From` for the mandatory source error                                                                                                        |
 | `context(suffix(N))`            | Changes the suffix of the generated context selector to `N`                                                                                                                                     |
 | `context(suffix(false))`        | No suffix for the generated context selector                                                                                                                                                    |
+| `transparent`                   | Delegates `Display` and `Error::source` to this error's source, implies `context(false)`)                                                                                                       |
```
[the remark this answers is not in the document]
seritools: was already removed in the fixup

tests/transparent.rs
seritools: Neither, just used no_context.rs as base: https://github.com/shepmaster/snafu/blob/19f667a80a0d2b5474ac1dc1b852f7cbea96d1ec/tests/no_context.rs#L45-L86

seritools: Just realized why I chose that particular one - the rules for when `snafu(transparent)` is allowed are the same as `snafu(context(false))`, since it is implied by `transparent`.

src/Snafu.md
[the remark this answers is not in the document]
shepmaster: Oops; I thought I switched out of the per-commit view. 🙇

snafu-derive/src/lib.rs
seritools: Just added it the last commit

snafu-derive/src/shared.rs
seritools: It's now tested by the compile-fail test, if that's what you mean. :)

seritools:
I think everything should be fixed up now - rewrote the docs and added an example.
Added another validation for `transparent NAND display(...)`.

shepmaster:
> how different is this implementation from what we do for an opaque error? Should there be some sort of unification?

TL;DR: Probably, but not too worried about it right now. We will want to revisit this when we support enums with tuple variants.

These are basically the same:

```rust
struct MyError(YourError);

#[snafu(transparent)]
struct MyError { source: YourError }
```

I think we should make these two the same:

```rust
enum MyError {
    MyError(YourError),
}

enum MyError {
    #[snafu(context(false))]
    MyError { source: YourError },
}
```

However, that would mean that our opaque errors, which are tuple structs / `struct MyError(YourError)`, would differ from tuple struct *enum variants* by the `transparent`. We should figure out how to perform some sort of reasonable migration.

shepmaster (approved):
[let it go](https://www.youtube.com/watch?v=FnpJBkAMk44)

PR #413, opened 2023-11-28
Replace doc-comment crate with language functionality

shepmaster:


shepmaster:
First build error was fixed by #416. Next should be fixed by #414

PR #416, opened 2023-11-28
Target Rust 1.61 by default

shepmaster:
Rust 1.61 came out on 2022-05-19, so it's safe enough to treat as available by default. Rust 1.56 continues to be supported by removing the default feature flags.


Enet4 (approved):
Looks good to me. :+1:

PR #417, opened 2023-11-28
Remove source from default Display implementation 

shepmaster:


Enet4 (approved):
There is great anticipation in these changes, which finally make the default behavior compliant with the recommendations from the error handling WG. :+1: :rocket:

---

Most of the changes are replacements in favor of implicit named arguments in all examples and tests that are sure to be run in a v1.61+ toolchain. The compatibility tests for v1.56 stay unchanged, so it's fine.

PR #426, opened 2023-12-04
Use `std::backtrace::Backtrace` as the default backtrace provider

### LINKED ISSUE #357: Use the standard library backtrace type as the default backtrace provider
I assume we will want to continue supporting the backtrace crate for a while to support older Rust versions, but we should be able to change our default to the standard library to reduce compile times for most up-to-date users.

shepmaster:
This also bumps our default Rust version to 1.65.

Closes #357


Enet4 (approved):
Looks solid to me. `std::backtrace::Backtrace` by default is also a welcoming addition. :+1:

PR #427, opened 2023-12-04
Add aliases for `[snafu(foo)]` to the `Snafu` macro

shepmaster:
This allows a search for `backtrace` to resolve to the macro, hinting where it is documented.


Enet4 (approved):
An interesting feature. :+1:

Screenshot 2023-12-05 092157

PR #448, opened 2024-04-02
Make Whatever thread-safe by default (add markers `Send + Sync`)

### LINKED ISSUE #446: Why isn't Whatever Send + Sync by default?
Hi @shepmaster!  Thank you for the wonderful crate!

### Problem

I have been switching my project to utilize `snafu`.  Previously, I used `anyhow::Error` as a generic error type when I didn't understand how to implement error types.  Now I am rewriting my project with proper error types, and in the meantime switching out `anyhow::Error` with `snafu::Whatever`.

When switching, I ran into a problem.  I have a future being spawned via Tokio, where the future has a output type of `Result<(), snafu::Whatever>`.  As this is concurrency, the rust compiler complains that that `Whatever` does not have `Send + Sync` markers for its source type.   The compile error is extremely indirect about this issue.  I have the error included at the bottom of this post.  To figure out where this issue was coming from required a bunch of research, as shown below.

### Research

Searching "snafu thread saftey" on Google returns the project `http-whatever`, where they essentially implement a new Whatever but with `Send + Sync` attached to the `StdError`:

https://github.com/bassmanitram/http-whatever/blob/af4fb2d672011e76a5746c27f31a2fdf65979326/src/lib.rs#L84-L93

Snafu's Whatever:

https://github.com/shepmaster/snafu/blob/073cc389dbdab8ac0dba7497377638f1170b7f89/src/lib.rs#L1557-L1569

I found a [snafu issue from January 2022](https://github.com/shepmaster/snafu/issues/322) that asks for implementation details for making a custom snafu error type be async compatible.  From there in May 2022 a [PR was merged](https://github.com/shepmaster/snafu/pull/331/files) that showed an error type with async/multi-threading as part of the tests.  This is part of the tests, but not part of the examples, so finding this example required a bunch of searching as its not part of the doc examples.

In addition, `anyhow::Error` has `Sync + Send` markers ([source in anyhow code](https://github.com/dtolnay/anyhow/blob/9f84a37551de7e7d67dc27a6e1c8518aa21f1d93/src/error.rs#L32), [source in anyhow docs](https://docs.rs/anyhow/latest/anyhow/struct.Error.html)).

### Question

This brings up the main question, why isn't `Whatever` Send + Sync by default?  Sure, I could implement a new Whatever as done by `http-whatever` and as shown in the snafu test.  But, if we consider `snafu::Whatever` to be a catch-all error type until the user can make more specific types, then I would expect as a user that it can be used in all places including async and multi-threading.

Thank you!

---

### Whatever Send + Sync error report
<details open>

```
error[E0277]: `(dyn StdError + 'static)` cannot be sent between threads safely
    --> src\file\file.rs:435:88
     |
435  |   ...                   task = Some(Handle::current().spawn(async move {
     |  ____________________________________________________________________-----_^
     | |                                                                    |
     | |                                                                    required by a bound introduced by this call
...    |
452  | | ...                       return Ok::<(), snafu::Whatever>(());
453  | | ...                   }));
     | |_______________________^ `(dyn StdError + 'static)` cannot be sent between threads safely
     |
     = help: the trait `std::marker::Send` is not implemented for `(dyn StdError + 'static)`
     = note: required for `Unique<(dyn StdError + 'static)>` to implement `std::marker::Send`
note: required because it appears within the type `Box<(dyn StdError + 'static)>`
    --> C:\Users\zardini123\.rustup\toolchains\1.76.0-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\alloc\src\boxed.rs:195:12
     |
195  | pub struct Box<
     |            ^^^
note: required because it appears within the type `std::option::Option<Box<(dyn StdError + 'static)>>`
    --> C:\Users\zardini123\.rustup\toolchains\1.76.0-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\core\src\option.rs:570:10
     |
570  | pub enum Option<T> {
     |          ^^^^^^
note: required because it appears within the type `Whatever`
    --> C:\Users\zardini123\.cargo\registry\src\index.crates.io-6f17d22bba15001f\snafu-0.8.2\src\lib.rs:1563:12
     |
1563 | pub struct Whatever {
     |            ^^^^^^^^
note: required because it appears within the type `Result<(), Whatever>`
    --> C:\Users\zardini123\.rustup\toolchains\1.76.0-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\core\src\result.rs:502:10
     |
502  | pub enum Result<T, E> {
     |          ^^^^^^
note: required by a bound in `Handle::spawn`
    --> C:\Users\zardini123\.cargo\registry\src\index.crates.io-6f17d22bba15001f\tokio-1.36.0\src\runtime\handle.rs:189:20
     |
186  |     pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
     |            ----- required by a bound in this associated function
...
189  |         F::Output: Send + 'static,
     |                    ^^^^ required by this bound in `Handle::spawn`
```

</details>

zardini123:
Allows Whatever's to be sent between threads by making them thread-safe by marking the source `std::error::Error` with `Send + Sync`.

For more discussion on why, see issue #446.

Resolves #446.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1561,9 +1561,9 @@ macro_rules! location {
 #[snafu(provide(opt, ref, chain, dyn std::error::Error => source.as_deref()))]
 #[cfg(any(feature = "std", test))]
 pub struct Whatever {
-    #[snafu(source(from(Box<dyn std::error::Error>, Some)))]
+    #[snafu(source(from(Box<dyn std::error::Error + Send + Sync>, Some)))]
     #[snafu(provide(false))]
-    source: Option<Box<dyn std::error::Error>>,
+    source: Option<Box<dyn std::error::Error + Send + Sync>>,
     message: String,
     backtrace: Backtrace,
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1558,7 +1558,7 @@ macro_rules! location {
 #[snafu(crate_root(crate))]
 #[snafu(whatever)]
 #[snafu(display("{message}"))]
-#[snafu(provide(opt, ref, chain, dyn std::error::Error => source.as_deref()))]
+#[snafu(provide(opt, ref, chain, dyn std::error::Error + Send + Sync => source.as_deref()))]
 #[cfg(any(feature = "std", test))]
 pub struct Whatever {
     #[snafu(source(from(Box<dyn std::error::Error + Send + Sync>, Some)))]
--- a/tests/backtrace.rs
+++ b/tests/backtrace.rs
@@ -130,7 +130,7 @@ mod whatever_nested {
         not_a_whatever().with_whatever_context(|_| format!("Outer failure"))
     }
 
-    fn not_a_whatever() -> Result<(), Box<dyn std::error::Error>> {
+    fn not_a_whatever() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         inner_whatever().map_err(Into::into)
     }
 


failable:
News?

Enet4:
Well, since my comment on https://github.com/shepmaster/snafu/issues/446#issuecomment-2033809122, no one has expressed concern over this change, but this also doesn't mean that someone out there is depending on `Whatever` not having to be `Send` or `Sync`.

At some point I hope we can move forward with this, since it does seem desirable and makes the error type more usable. It will imply a semver bump in any case.

Stargateur:
Still think sync is not needed

failable:
So can you share the best practices when using `snafu` in async Rust? Thanks!

PR #451, opened 2024-04-13
[snafu-derive] Update heck to v0.5

Enet4:
No further changes appear to be necessary.


shepmaster:
Away from computer at the moment, but I think I want this to be `default-features=false`, have the version range be `>= 0.4, < 0.6`, and then probably add a test that it continues to work with version 0.4. 0.5 will be tested in CI automatically.

shepmaster:
> and then probably add a test

We [already have that](https://github.com/shepmaster/snafu/blob/36c6bbe3ead565b4c9aa65e5f748aad43bdfb89a/.cirrus.yml#L133), yay.

shepmaster (approved):

sunshowers:
Hi there -- as you might be aware, there's a long-standing issue with Cargo (https://github.com/rust-lang/cargo/issues/5702) where specifying a too-wide version of a dependency can cause cargo updates to flap around between the old and the new versions. Would you be okay just saying heck 0.5?

shepmaster:
> as you might be aware

I was not, so thanks for the heads up!

> a long-standing issue with Cargo 

I actually thought about fixing the bug in Cargo instead of working around it in my crate. However, the reproduction case in that issue doesn’t work anymore. Since you presumably saw this recently, do you have a way to reliably reproduce the problem?

sunshowers:
Sure -- we noticed it in one of our (Oxide's) larger repos:

```
git clone https://github.com/oxidecomputer/omicron
cd omicron
git checkout 38cb5542
cargo update -p clap --precise 4.5.21
```

This produces this diff: https://gist.github.com/sunshowers/2973e27d91f17d1a92f6a1fe026ee79c

As you can see, the snafu-derive -> heck dependency got updated from 0.5 to 0.4. (I've also observed it go the other way.)

PR #455, opened 2024-05-25
Add boxed_context and with_boxed_context

ChristopherRabotin:
Hi there,

First off, thanks for the work on snafu. I started using it a few months ago, and it's fantastic.

I've used it so thoroughly, that I'm now encountering warnings in a pretty large library I maintain where the error variant is getting quite large. As such, the compiler recommends that I return boxed errors, i.e. `Result<T, Box<E>>` instead of `Result<T, E>`.

`snafu` already provides a `Box<dyn Error>` return capability. Rightfully so, the documentation discourages its use. As such, I'm proposing the addition of `boxed_context` and `with_boxed_context`: they work as `context` and `with_context` respectively. 

Let me know if this is a fringe use case, and if so, what is the recommendation for returning a boxed error via `with_context`.

Thanks

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -566,6 +566,45 @@ pub trait ResultExt<T, E>: Sized {
         C: IntoError<E2, Source = E>,
         E2: Error + ErrorCompat;
 
+    /// Extend a [`Result`]'s error with additional context-sensitive information, returned in Box to prevent large Enum variants.
+    ///
+    /// [`Result`]: std::result::Result
+    ///
+    /// ```rust
+    /// use snafu::prelude::*;
+    ///
+    /// #[derive(Debug, Snafu)]
+    /// enum Error {
+    ///     Authenticating {
+    ///         user_name: String,
+    ///         user_id: i32,
+    ///         source: ApiError,
+    ///     },
+    /// }
+    ///
+    /// fn example() -> Result<(), Error> {
+    ///     another_function().context(AuthenticatingSnafu {
+    ///         user_name: "admin",
+    ///         user_id: 42,
+    ///     })?;
+    ///     Ok(())
+    /// }
+    ///
+    /// # type ApiError = Box<dyn std::error::Error>;
+    /// fn another_function() -> Result<i32, ApiError> {
+    ///     /* ... */
+    /// # Ok(42)
+    /// }
+    /// ```
+    ///
+    /// Note that the context selector will call [`Into::into`][] on each field,
+    /// so the types are not required to exactly match.
+    #[cfg(feature = "std")]
+    fn boxed_context<C, E2>(self, context: C) -> Result<T, Box<E2>>
+    where
+        C: IntoError<E2, Source = E>,
+        E2: Error + ErrorCompat;
+
     /// Extend a [`Result`][]'s error with lazily-generated context-sensitive information.
     ///
     /// [`Result`]: std::result::Result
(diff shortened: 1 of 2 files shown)


shepmaster:
> As such, the compiler recommends

Presumably you mean [Clippy's `result_large_err`](https://rust-lang.github.io/rust-clippy/stable/index.html#/result_large_err), right? 

```rust
use snafu::prelude::*;

fn main() {
    println!("{:?}", usage());
}

fn usage() -> Result<(), Error> {
    TinySnafu.fail()
}

#[derive(Debug, Snafu)]
enum Error {
    Tiny,

    Big { source: BigSourceError },
}

#[derive(Debug, Snafu)]
struct BigSourceError {
    data: [u8; 1024],
}
```

```
warning: the `Err`-variant returned from this function is very large
  --> src/main.rs:7:15
   |
7  | fn usage() -> Result<(), Error> {
   |               ^^^^^^^^^^^^^^^^^
...
15 |     Big { source: BigSourceError },
   |     ------------------------------ the largest variant contains at least 1024 bytes
   |
   = help: try reducing the size of `Error`, for example by boxing large elements or replacing it with `Box<Error>`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
   = note: `#[warn(clippy::result_large_err)]` on by default
```

In these cases, I usually [transform the source](https://docs.rs/snafu/latest/snafu/derive.Snafu.html#transforming-the-source) to box it:

```diff
diff --git a/src/main.rs b/src/main.rs
index f0fd409..ff73c54 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,10 @@ fn usage() -> Result<(), Error> {
 enum Error {
     Tiny,
 
-    Big { source: BigSourceError },
+    Big {
+        #[snafu(source(from(BigSourceError, Box::new)))]
+        source: Box<BigSourceError>,
+    },
 }
 
 #[derive(Debug, Snafu)]
```

This has the nice benefit that any existing `.context` calls will automatically perform the boxing. Do you think that would work in your case?

ChristopherRabotin:
Excellent, yes, I'm sure this will indeed solve my use case, thank you!

ChristopherRabotin:
I'm only trying this out now, and I'm seemingly not using it correctly, but I can't seem to figure out what I'm doing wrong.

Here is what I did:

```diff
--- a/src/od/mod.rs
+++ b/src/od/mod.rs
@@ -158,11 +158,21 @@ where
     }
 }

+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(crate)))]
+pub enum Error {
+    SourceNeedsToBeBoxed {
+        #[snafu(source(from(TrajError, Box::new)))]
+        source: Box<TrajError>,
+    },
+}
+
 #[derive(Debug, PartialEq, Snafu)]
 #[snafu(visibility(pub(crate)))]
 pub enum ODError {
-    #[snafu(display("during an orbit determination, encountered {source}"))]
-    ODPropError { source: PropagationError },
+    // #[snafu(display("during an orbit determination, encountered {source}"))]
+    #[snafu(source(from(PropagationError, Box::new)))]
+    ODPropError { source: Box<PropagationError> },
     #[snafu(display("during an orbit determination, encountered {source}"))]
     ODDynamicsError { source: DynamicsError },
     #[snafu(display("at least {need} measurements required for {action}"
```

And here is the error I'm getting:
```txt
error: `source` attribute is only valid on enum variant or struct fields with a name, not on an enum variant
   --> src/od/mod.rs:174:13
    |
174 |     #[snafu(source(from(PropagationError, Box::new)))]
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

Note that the error only happens in the `ODError` enum, not in the `Error`, which is from the [documentation](https://docs.rs/snafu/latest/snafu/derive.Snafu.html#transforming-the-source) you linked to.

Further, I haven't figured out how to add a `display` to a Boxed error. It seems to also cause the same error.

What am I misunderstanding here?

Thanks

Stargateur:
there is a problem with the error message "error: `source` attribute is only valid on enum variant or struct fields with a name, not on an enum variant"

> "only valid on enum variant"
> "not on an enum variant"

XD I think you want to write "is only valid on enum variant field" or something like that. I think the best would be to write "error: `source` attribute is only valid for fields, not on an enum variant"

ChristopherRabotin:
I was confused by the error because, as Stargateur pointed out, it said that I need to place it on the enum variant ... which is what I was doing. I don't know how I didn't see the difference between the demo code and what I coded, but it's obvious now, and it works out of the box.

Thank you!

alerque:
> > error: `source` attribute is only valid on enum variant or struct fields with a name, not on an enum variant
> 
> I don't know how to say it better than this message, so I'm really hoping you can help me reword it to be clearer!

Depending on what exactly you mean (I'm not 100% clear myself) perhaps something like one of these:

* "is not valid on enum variant or struct fields that do not have a name"
* "is only valid on named enum variants or struct fields, not on unnamed enum variants"
* "is only valid applied directly to enum variants or struct fields, not on the whole enum or struct"

PR #456, opened 2024-06-25
Use full path for `location!()` macro

discord9:
use full path of `snafu::Location` for `location!()` macro


shepmaster (approved):
Thank you!

PR #463, opened 2024-08-26
Support core::error in Rust 1.81

shepmaster:


src/lib.rs
```diff
@@ -338,25 +338,41 @@ generate_guide! {
...
 mod no_std_error;
-#[cfg(not(any(feature = "unstable-core-error", feature = "std", test)))]
+#[cfg(not(any(
+    feature = "rust_1_81",
+    feature = "unstable-core-error",
+    feature = "std",
+    test
+)))]
 #[doc(hidden)]
 pub use no_std_error::Error;
```
Enet4, line 377: Took me a minute, but IIUC this is the fallback `Error` provided when no other `Error` "offering" is available. A very minor thing, but the name `no_std_error` could at first mislead into thinking that the module holds important Snafu stuff around supporting `core::error::Error`, rather than a fallback, simpler `Error` trait. Would `fallback_error` convey this idea better than `no_std_error`?

Enet4 (approved):
Looks OK to me. I just left a comment on what could eventually be a very tiny misinterpretation of the internal code, but I will not block the PR over it.

ognevny:
what's progress with this?

PR #466, opened 2024-10-31
Add top-level documentation on reporting errors

### LINKED ISSUE #464: Mention Report in the top level docs
> i feel like [using `Report`] should be mentioned in the root crate documentation because i never knew that existed

Enet4:
Resolves #464. I'm open to the possibility of shortening the example code.


alerque (commented):
Add a potential consumer of the docs this shows me roughly what i could do but it is not at all apparent from the docs what the result output would be or why i would want to use this over done alternative.

Enet4:
@alerque Please see whether the new version of the documentation explains the situation better. At this point it is a matter of striking a balance between explaining the purpose of this mechanism and not becoming too redundant with the contents in `macro@report` and `Report`.

alerque (approved):
I think that is *much* better!

shepmaster (approved):
Looks good! I just squashed the commits and reworded the commit message a bit.

PR #482, opened 2025-05-02
Add #[allow(dead_code)] to the context selector's fail method

ldseraph:


shepmaster:
Thanks! However, I don't understand what problem this solves. Specifically, if I compile this with the currently released version of SNAFU, there are no warnings / errors:

```rust
#![deny(dead_code)]

use snafu::prelude::*;

#[derive(Debug, Snafu)]
struct MyError;

fn some_usage() -> Result<(), MyError> {
    Err(MyError)
}

fn main() {
    some_usage().unwrap_err();
}
```

Please help me understand what is reporting that the `fail` method is dead code.

ldseraph:
If I define an error within a macro, then a warning saying "the method 'fail' is never used" will occur.

```rust
pub mod error;

#[macro_export]
macro_rules! define_error {
    ({
        $(
            $(#[$variant_attr:meta])*
            $variant:ident $({
                $(
                    $(#[$attr:meta])*
                    $field:ident: $ty:ty
                ),* $(,)?
            })?
        ),* $(,)?
    }) => {
        use $crate::error::Snafu;
        pub type Result<T, E = Error> = core::result::Result<T, E>;
        #[derive(Debug, Snafu)]
        #[snafu(visibility(pub(crate)))]
        pub enum Error {
            $(
                $(#[$variant_attr])*
                $variant $({
                    $(
                        $(#[$attr])*
                        $field: $ty
                    ),*
                })?
            ),*
        }
    };
}

define_error!({
    #[snafu(display("Some"))]
    Some,
});

pub fn test() -> Result<()> {
    Err(SomeSnafu {}.build())
}

```

```bash
$ cargo check --all
warning: method `fail` is never used
  --> snafu-ext\src\lib.rs:18:25
   |
18 |           #[derive(Debug, Snafu)]
   |                           ^^^^^ method in this implementation
...
34 | / define_error!({
35 | |     #[snafu(display("Some"))]
36 | |     Some,
37 | | });
   | |__- in this macro invocation
   |
   = note: `#[warn(dead_code)]` on by default
   = note: this warning originates in the macro `define_error` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `snafu-ext` (lib) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
```

shepmaster:
Thanks for the explanation! I've opened rust-lang/rust#141005 as I don't think the behavior is correct.

As a side note, instead of writing`Err(SomeSnafu {}.build())`, you should write `Err(SomeSnafu.build())` (no reason for the braces) and then you should write `SomeSnafu.fail()` (that's the reason that `fail` exists).

shepmaster:
Thanks!

PR #490, opened 2025-05-13
Document cfg's on docs.rs

jplatte:
Preview using `RUSTDOCFLAGS='--cfg=docsrs' cargo +nightly doc --features futures,guide --open`:

Screenshot 2025-05-13 at 22-43-46 snafu - Rust

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,6 @@
 #![deny(missing_docs)]
 #![allow(stable_features)]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![cfg_attr(not(any(feature = "std", test)), no_std)]
 #![cfg_attr(feature = "unstable-core-error", feature(error_in_core))]
 #![cfg_attr(


shepmaster:
Thanks; this is something I've been meaning to get around to and just kept putting off!

> I saw a lot of rustdoc warnings locally when generating the docs. I can send another PR if you're interested :)

Yeah, that'd be nice. I just saw those myself when doing a bit of SNAFU work today. The only thing is that I'm not sure if some of those warnings were because of supporting older Rust versions.

> Netlify preview doesn't show the update because it doesn't use `--cfg docsrs`

I'd actually be fine with it being updated to do that.

src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![deny(missing_docs)]
 #![allow(stable_features)]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```
shepmaster, line 3: Hah, I didn't even know about `doc_auto_cfg`, I was planning on doing this the Very Hard Way 😅

shepmaster:
Thanks!

PR #491, opened 2025-05-15
a typo in philosophy.md

sify21:
`bin` -> `bind`


Enet4 (changes requested):
This isn't really a typo, as described in #185.

> This isn't actually a typo. I'm using "bin" as a verb to mean "put into a box" / "sort" / "categorize". Can you think of a less confusing way of stating the same concept?

The verb _"bind"_ does not seem to portray the intended concept either.

PR #492, opened 2025-05-15
Bring netlify preview closer to docs.rs

jplatte:
Doc references to non-sysroot crates are also going to work in the Netlify preview with this. For example: The `TryFuture` and `TryStream` links in the `futures` module.

This does not impact development with non-nightly toolchains, since they conveniently just ignore unstable settings in `.cargo/config.toml` (they do not ignore the same settings if passed via environment variable or CLI flag, IIRC).

Still, if you don't want to commit `.cargo/config.toml`, the same thing could probably be done via environment variables within `netlify.toml`.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,7 +1,7 @@
 [build]
 command = """
 rustup install nightly --profile minimal && \
-RUSTDOCFLAGS='--cfg=docsrs' cargo +nightly doc --no-deps --features=futures,guide
+RUSTDOCFLAGS='--cfg=docsrs' cargo +nightly doc --no-deps --features=futures,guide -Zrustdoc-map
 """
 publish = "target/doc"
 


shepmaster:
Thanks!

You may want to update / rebase on main as it now has your previous changes. I was confused as to why the nice `cfg` symbols disappeared in this preview 😉 

> the same thing could probably be done via environment variables within `netlify.toml`.

Yeah, my instinctual thought was "oh no, not `config.toml`". Would you check to see if there are environment variables we can set inside of there? If not, I think we could use the shell script to write the `config.toml`.

jplatte:
Looks like just enabling the unstable feature is sufficient, no need for the extra bit of config I copy-pasted from another project where this sort of thing is done :)

shepmaster (approved):

shepmaster:
Thanks!

PR #499, opened 2025-05-28
Switch to unconditional #![no_std]

jplatte:
… with conditional extern crate std.
This disables the default prelude and thus makes it easier to notice when developing when one adds a dependency on alloc / std functionality.


jplatte:
Pretty sure that this now allows using the `futures` feature w/o the `std` feature which was broken before :)

PR #509, opened 2025-08-14
Implement GenerateImplicitData and AsBacktrace for various boxes

shepmaster:


Enet4 (approved):
Clear and sound! Also peculiar to see a module dedicated with nothing but impls, which is an interesting pattern that still suits this purpose well.

PR #534, opened 2025-12-05
Update MSRV to Rust 1.65

shepmaster:


shepmaster:
> I'm sure you probably know this

I didn't, actually, so thank you! However, I was coincidently planning on stopping the MSRV update train here at 1.65. Targeting a ~3 year old Rust compiler has been somewhat of the sweet spot, I think.

That said, I'm not sure if or how I'll let this knowledge influence my decisions going forward. *Presumably* any Ubuntu-provided Rust packages would continue using SNAFU 0.8.x as the packages wouldn't get updated to use a new version anyway. That leaves "people compiling Rust code themselves on Ubuntu but using the Ubuntu-provided Rust compiler". That group is able to install a newer Rust compiler (I assume this without checking) if they are developing their own projects. If they are just compiling some other project that uses SNAFU, then that project would be responsible for keeping an old-enough SNAFU version.

Thankfully, [Rust 1.84](https://blog.rust-lang.org/2025/01/09/Rust-1.84.0/#cargo-considers-rust-versions-for-dependency-version-selection) will make this a lot easier going forward; we just have to wait for that to be in all the LTS streams. 😪

alerque:
> _Presumably_ any Ubuntu-provided Rust packages would continue using SNAFU 0.8.x as the packages wouldn't get updated to use a new version anyway. That leaves "people compiling Rust code themselves on Ubuntu but using the Ubuntu-provided Rust compiler". That group is able to install a newer Rust compiler (I assume this without checking) if they are developing their own projects.

This reasoning covers the vast majority of cases, but there is a 3rd case. You're right about how things work for the first 2 major cases — the third case is 3rd party PPAs that specifically target older systems that for whatever policy reasons are not getting updated until the end of LTS cycles but need *some* fresh software. There are quite a few cases out there in production using mostly ancient system packages and then putting a handful of vetted mission specific builds in a PPA. Of course it is *possible* to also build these with a newer toolchain, but that introduces quite a few quirks and potentially more things to get special approval for. It's often easier to stick with the as many system components as possible. There is actually an updates channel published by Canonical that bumps the Bionic Rust package as far as ... 1.71 maybe? ... which helps a lot. Using `rustup` to build Ubuntu packages is actually moronically harder than it should be.

Of course as you said this doesn't have to hold this project back as apps that have to target older systems can also just hold back the libraries they use. I was just pointing out that as of 1.65 you were about to jump a big line in the sand. Maybe it's *worth* jumping, but at least know that past this point you'll start creating a lot more downstream friction. Circa 2025.

shepmaster:
> Thankfully, [Rust 1.84](https://blog.rust-lang.org/2025/01/09/Rust-1.84.0/#cargo-considers-rust-versions-for-dependency-version-selection) will make this a lot easier going forward

In addition, since we gained `rust-version` in Cargo.toml, Cargo will at least report things and guide towards a solution. For example:

```
error: package `quote v1.0.42` cannot be built because it requires rustc 1.68 or newer, while the currently active rustc version is 1.65.0
Either upgrade to rustc 1.68 or newer, or use
cargo update -p quote@1.0.42 --precise ver
where `ver` is the latest version of `quote` supporting rustc 1.65.0
```

PR #554, opened 2026-03-25
Use namespaced dependency activation

jplatte:
Stabilized in 1.60, which is lower than the current MSRV: https://blog.rust-lang.org/2022/04/07/Rust-1.60.0/#new-syntax-for-cargo-features


shepmaster (approved):

jplatte:
Ah yeah, I should have mentioned that I found no `cfg` use of the implicit features :)

PR #556, opened 2026-05-04
Expand Self to the name of the error type when generating selectors

### LINKED ISSUE #218: Using `Self` in an context selector ends up referring to the selector, not the error type
I'm unclear if this is considered a bug or not, but it is surprising!

```rust
enum Error {
    Example { trace: Vec<Self> },
}
```

legeana:
Self used inside a field type always refers to the error type itself. Previously such a type was copied verbatim to the *Snafu selector type, which changes the meaning of Self from the error type to the selector type. This change expands Self when capturing the field type.

The type expansion is implemented using a visit_type_path_mut provided by the syn crate.

Closes #218.

### DIFF
--- a/snafu-derive/Cargo.toml
+++ b/snafu-derive/Cargo.toml
@@ -32,4 +32,5 @@ features = [
     "parsing",
     "printing",
     "proc-macro",
+    "visit-mut",
 ]
--- a/snafu-derive/src/lib.rs
+++ b/snafu-derive/src/lib.rs
@@ -290,8 +290,50 @@ fn impl_snafu_macro(ty: syn::DeriveInput) -> TokenStream {
     }
 }
 
+const SELF_TYPE: &str = "Self";
+
+struct ExpandSelfVisitor {
+    self_ty: syn::TypePath,
+}
+
+impl syn::visit_mut::VisitMut for ExpandSelfVisitor {
+    fn visit_type_path_mut(&mut self, node: &mut syn::TypePath) {
+        // Terminate on `Self` nodes.
+        if node.qself.is_none() && node.path.is_ident(SELF_TYPE) {
+            *node = self.self_ty.clone();
+            return;
+        }
+        // Continue recursive visiting.
+        syn::visit_mut::visit_type_path_mut(self, node);
+    }
+}
+
+/// Returns a `DeriveInput` with all occurrences of `Self` replaced
+/// with its name.
+///
+/// We use pieces of the original code to create context selector
+/// types and implementations. If `Self` is copied verbatim, it will
+/// change its meaning from the error type the user provided to the
+/// context selector. Instead, we eagerly resolve `Self` to the name
+/// of the original type.
+fn expand_self(mut input: syn::DeriveInput) -> syn::DeriveInput {
+    let mut v = ExpandSelfVisitor {
+        self_ty: syn::TypePath {
+            qself: None,
+            // We don't need to worry about fully qualifying the original type
+            // name because we either refer to it from the same module, or from
+            // a child module that imports everything from the parent module.
+            path: input.ident.clone().into(),
+        },
+    };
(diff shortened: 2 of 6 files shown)


legeana:
Converting this to a draft because this change depends on #557. Once #557 is merged only the last commit will be needed.

shepmaster:
This all looks great, thank you! The only question I have is spurred by your comment:

```rust
    // Moving the value allows the call site to have an immutable syn_field.
```

I know that we pass around immutable references almost everywhere in the parsing code, but *way* up at the top, in `parse_snafu_information`, we have an owned `DeriveInput`:

https://github.com/shepmaster/snafu/blob/1f8e75f56390c421a198871916100c6316d23d4f/snafu-derive/src/lib.rs#L293

What would you think the code would look like if we moved the visitor invocation allllllll the way up there? Would that apply too broadly and convert things we wouldn't want converted?

shepmaster:
> if we moved the visitor invocation allllllll the way up there

That might also cover more esoteric issues like using `Self` in a trait bound:

```rust
    #[derive(Debug, PartialEq, Snafu)]
    struct Error
        where i32: From<Self>,
    {
        foo: i32,
    }
```

legeana:
> I know that we pass around immutable references almost everywhere in the parsing code, but way up at the top, in parse_snafu_information, we have an owned DeriveInput:

We could save one clone per field by doing this, yes. I don't expect this to have any performance impact though.

> What would you think the code would look like if we moved the visitor invocation allllllll the way up there? Would that apply too broadly and convert things we wouldn't want converted?

Let me try to work it out. The original implementation only targeted fields, so it was rather easy to reason about:
- In a field definition a TypePath is obviously a component of the field path.
- It is less clear (due to my limited AST knowledge) where a TypePath may appear within an entire enum/struct. It is probably safe to assume that would be named type external to the struct/enum definition.
- ~~My only concern replacing Self indiscriminately would be some unexpected interactions with linters, if they read the returned code. For example `clippy::use_self`. I am trying to check this now.~~
  - [is_from_proc_macro](https://doc.rust-lang.org/clippy/development/macro_expansions.html#the-is_from_proc_macro-function) may help prevent that.

> > if we moved the visitor invocation allllllll the way up there
> 
> That might also cover more esoteric issues like using `Self` in a trait bound:
> 
> ```rust
>     #[derive(Debug, PartialEq, Snafu)]
>     struct Error
>         where i32: From<Self>,
>     {
>         foo: i32,
>     }
> ```

Unfortunately I don't understand how your examples applies to Snafu specifically. Why do we care about the trait bound here?

The problem this PR addresses is `Self` in fields where the snafu-derive crate creates a new type that refers to these fields. Does Snafu also copy trait bounds from the original type? Is there a more complete example of this behaviour?

I would appreciate some help understanding the use case.
Thanks!

P.S. I uploaded a draft of what this would look like with a implementation at the root. I think we need more tests based on your example above.

legeana:
I believe I figured out what you meant in the example. Snafu propagates trait bounds to Snafu objects. Now this comment makes sense to me. Thanks for catching!

legeana:
I added a separate test for trait bounds. The test is rather simple so I combined both enum and struct cases in the same file.

The rest is basically the same as in the previous version of the code, except Self is now expanded at the top of the call hierarchy. Please take another look :)

legeana:
@shepmaster please take another look, in case notifications didn't work. I had issues with that in the past.

If you are busy no problem at all. Thanks!

shepmaster (approved):

shepmaster:
Thanks for your patience! This has been an annoying gotcha that's always been possible to work around, but not having to will be even better.

I ran [mini-crater](https://github.com/shepmaster/mini-crater) on this change:

| status           | count |
|------------------|------:|
| success          |   172 |
| error (original) |    15 |
| error            |     0 |

shepmaster:
And [released in 0.9.1](https://crates.io/crates/snafu/0.9.1)!

PR #557, opened 2026-05-05
Fix pre-submit failures for Rust 1.95.0

legeana:
The test was broken at main. This came up in #556.


shepmaster (approved):
Thank you for (a) splitting this out into a separate PR and (b) splitting the logical changes into separate commits! ❤️

PR #561, opened 2026-07-21
Migrate from CirrusCI to GitHub Actions

shepmaster:


shepmaster:
It's expected that this will fail as there are two upcoming changes needed to get back to green. I manually tested with them all combined over in https://github.com/shepmaster/snafu/actions/runs/29852638040.

PR #562, opened 2026-07-21
Update backtrace test for symbol mangling v0

shepmaster:


shepmaster:
It's expected that this does not pass CI, as we need yet another fix to get all the way to green. Tested all rolled together over in https://github.com/shepmaster/snafu/actions/runs/29852638040.

